PR #14, opened 2018-03-07
Add eslint

meriadec:
Adding airbnb eslint config, disabling annoying rules, fixing majority of issues.
I've intentionnaly let those ones:

```
react/no-did-mount-set-state:  1| 
react/no-string-refs:          2|  
react/no-unused-state:        15|               
react/prop-types:             11|           
```

because of not sure about what to do with thoses states fields, string refs, etc.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,9 @@
-//@flow
+// @flow
+
 import "./polyfill";
-import React, { Component } from "react";
-import { translate } from "react-i18next";
-import { NativeModules } from "react-native";
+
+import React, { Component } from "react"; // eslint-disable-line import/first
+
 import App from "./App";
 import { LocaleProvider } from "./components/LocaleContext";
 
--- a/src/screens/ReceiveFunds.js
+++ b/src/screens/ReceiveFunds.js
@@ -1,5 +1,6 @@
 /* @flow */
 import React, { Component } from "react";
+import AppBtc from "@ledgerhq/hw-app-btc";
 import {
   View,
   ScrollView,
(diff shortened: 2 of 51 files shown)


src/index.js
```diff
@@ -1,8 +1,6 @@
-//@flow
-import "./polyfill";
+// @flow
 import React, { Component } from "react";
-import { translate } from "react-i18next";
-import { NativeModules } from "react-native";
+import "./polyfill";
```
gre: it is important the polyfill is the first thing we import. it might lead to undesired problem otherwise.

src/locales/index.js
```diff
@@ -1,2 +1,2 @@
-exports.en = { common: require("./en/common.json"), };
-exports.fr = { common: require("./fr/common.json"), };
+exports.en = { common: require("./en/common.json") };
+exports.fr = { common: require("./fr/common.json") };
```
gre: this file is generated, we don't want to edit it

src/screens/ReceiveFunds.js
```diff
@@ -45,7 +39,7 @@ export default class ReceiveFunds extends Component<*, *> {
 
   componentDidMount() {
     const { params } = this.props.navigation.state;
-    let amount = 0;
+    const amount = 0;
```
gre: I think there is a bug here, the internal `let amount` shouldn't be a variable but the same. anyway this part is likely temporary

src/locales/index.js
meriadec: ah :) let make prettier ignore it

src/screens/ReceiveFunds.js
meriadec: not sure I get what you mean. do you think it should be a `let` ?

src/index.js
meriadec: updated (even if I don't think importing after the `react` import will break something) :)

src/screens/ReceiveFunds.js
gre: that's ok. there is just a broken logic in the code. that whole block is literally useless https://github.com/LedgerHQ/ledger-wallet-mobile/blob/master/src/screens/ReceiveFunds.js#L49-L54 because accidentally have an inner let amount, where it should just be an assignment, not a declaration

gre:
#15

PR #22, opened 2018-03-21
Account settings

NastiaS:


src/components/UnitRow.js
```diff
@@ -0,0 +1,40 @@
...
+  unit: Unit,
+  onPress: () => void
+}> {
+  render() {
+    const { unit, onPress } = this.props;
+    return (
+      <TouchableOpacity onPress={onPress}>
+        <View
+          style={{
+            minHeight: 50,
```
gre: you can use a StyleSheet as well like you did for SectionTitle.

src/mock/account.js
```diff
@@ -99,6 +99,7 @@ export function genAccount(accountIndex: number) {
     balance: 0,
     address: genAddress(currency),
     operations: [],
+    minConfirmations: 0,
```
gre: let's set it to 2 in mock because 0 would not be so interesting

src/screens/accountSettings/EditConfirmations.js
```diff
@@ -0,0 +1,75 @@
...
+    goBack: () => void
+  }>,
+  updateAccount: ($Shape<Account>) => *
+}> {
+  static navigationOptions = {
+    title: "Confirmations"
+  };
+  onConfNumEndEditing = (e: *) => {
+    const { account } = this.props.navigation.state.params;
+    e.nativeEvent.text.length
```
gre: as the result is not used in this ternary instruction, I would refer it to be rewritten to

```
if (e.nativeEvent.text.length) {
   updateAccount(...)
}
```

src/screens/accountSettings/EditConfirmations.js
```diff
@@ -0,0 +1,75 @@
...
+      ? this.props.updateAccount({
+          minConfirmations: parseInt(e.nativeEvent.text),
```
gre: minor general best practice: it is recommended to use `parseInt(..., 10)` to make sure we use decimal parsing.

src/screens/accountSettings/EditUnits.js
```diff
@@ -0,0 +1,76 @@
...
+  updateAccount
+};
+
+class EditUnits extends Component<{
+  updateAccount: ($Shape<Account>) => *,
+  navigation: NavigationScreenProp<{
+    params: {
+      account: Account
+    },
+    goBack: () => void
```
gre: I don't think you need to define that. goBack() is not part of `this.props.navigation.state`, it is part of `this.props.navigation` which directly is typed by `NavigationScreenProp`. 
see https://github.com/react-navigation/react-navigation/blob/e6559f5878995e7fc9dd6ffe9036bbe3f584e54e/flow/react-navigation.js#L488

src/screens/accountSettings/EditUnits.js
```diff
@@ -0,0 +1,76 @@
...
+  }>
+}> {
+  static navigationOptions = {
+    title: "Edit Units"
+  };
+  onItemFullPress = (item: *) => {
+    const { account } = this.props.navigation.state.params;
+    const { navigation, updateAccount } = this.props;
```
gre: minor tiny simplification proposal:

```
    const { navigation, updateAccount } = this.props;
    const { account } = navigation.state.params;
```

PR #23, opened 2018-03-22
walletcom-71 receive funds clean up

NastiaS:


src/screens/ReceiveFundsSelectAccount.js
```diff
@@ -25,7 +25,8 @@ class ReceiveFundsSelectAccount extends Component<{
   navigation: NavigationScreenProp<{
     params: {
       selectedAccountId: string,
-      setAccountId: string => void
+      setAccountId: string => void,
+      syncCounterValue: string => void
```
gre: We shouldn't introduce another callback than just setAccountId. instead we should do the syncCounterValue logic on the other side.

gre: you can remove syncCounterValue from the prop type

src/screens/ReceiveFundsSelectAccount.js
```diff
@@ -36,13 +36,9 @@ class ReceiveFundsSelectAccount extends Component<{
...
-    const {
-      selectedAccountId,
-      setAccountId,
-      syncCounterValue
-    } = navigation.state.params;
-    if (selectedAccountId !== account.id) setAccountId(account.id);
-    syncCounterValue(account.unit.code);
+    const { selectedAccountId, setSelectedAccount } = navigation.state.params;
+    if (selectedAccountId !== account.id)
+      setSelectedAccount(account.id, account.unit.code);
```
gre: to keep things the most canonical possible, I think it should just be setting back the accountId like before, but only the other side would have to find the account by id again to find what it its code.

> On a side note, ideally, we should find a better way to communicate back to the parent navigation. the fact we give a callback in the navigate is still a hack to me but I have not found a better quick solution for it. It was suggested by some folks of react-navigation but I feel there are some design problem with it (a callback is not serializable / can't recreate the screen stack if things are connected with a cb). I would dream if it was possible to do `navigation.goBackWithAnArgument()` and the previous parent screen would handle it

(also, one other thing to note is that unit.code is not really reliable & we'll introduce currency.ticker later)

PR #31, opened 2018-03-28
WALLETCOM-68 Price evolution UI

MortalKastor:


src/components/PriceChange.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+import React, { Component } from "react";
+import LText from "./LText";
+import colors from "../colors";
+
+export default class PriceChange extends Component<{
+  before: number,
+  after: number,
+  color?: "western" | "eastern",
```
gre: instead of accepting color here, I think you should accept a `locale` prop, and we would use `withLocale()` to inject it. ( https://github.com/LedgerHQ/ledger-wallet-mobile/blob/master/src/components/LocaleContext.js#L60-L72 )

then we basically need to figure out how to infer color from that locale  🤔  . also would this be something in global settings to be able to override it?

src/components/PriceChange.js
```diff
@@ -0,0 +1,34 @@
...
+    const percent: number = after / before * 100 - 100;
+
+    if (color === "western") {
+      style.color = percent >= 0 ? colors.green : colors.red;
+    } else if (color === "eastern") {
+      style.color = percent >= 0 ? colors.red : colors.blue;
+    }
+
+    return (
+      <LText style={[this.props.style, style]}>
```
gre: lil' detail here, but reverse style & this.props.style IMO. (basically we should allow style to override color if needed)

gre: use `numberOfLines={1}` to make sure it does not wrap

src/components/PriceChange.js
```diff
@@ -0,0 +1,34 @@
...
+        {`${(percent >= 0 ? "+" : "") + percent.toFixed(2)} %`}
```
gre: according to desktop wireframes, we want a space between the sign and the number. so you might have to `?"+ ":"- "` and to `Math.abs(percent)..`

src/screens/Accounts.js
```diff
@@ -24,6 +24,7 @@ import CurrencyUnitValue from "../components/CurrencyUnitValue";
 import WhiteButton from "../components/WhiteButton";
 import CurrencyIcon from "../components/CurrencyIcon";
 import CounterValue from "../components/CounterValue";
+import PriceChange from "../components/PriceChange";
```
gre: My mistake here, but I think let's rename this to `DeltaChange`. because technically it's not about the price, component works with comparing anything.

src/screens/Accounts.js
```diff
@@ -93,7 +94,9 @@ class AccountCard extends PureComponent<*, *> {
   };
   render() {
     const { account, onPress } = this.props;
-    const data = getBalanceHistory(account, 30);
+    const data = getBalanceHistory(account, 10);
+    const startPrice: number = data[0].value;
+    const endPrice: number = data[data.length - 1].value;
```
gre: good practice here. we will refactor more things according to this idea everything is already in the balance history data 🚀

is it possible to not type `: number` here ? if it does infer properly, my general practice is to not type the local variables

src/screens/Accounts.js
```diff
@@ -169,6 +172,11 @@ class AccountCard extends PureComponent<*, *> {
                   currency={account.currency}
                 />
               </LText>
+              <PriceChange
+                before={startPrice}
+                after={endPrice}
+                color="western"
```
gre: don't provide it, as noted above

src/screens/Dashboard.js
```diff
@@ -47,19 +48,31 @@ class ListHeaderComponent extends PureComponent<
...
+              <CurrencyUnitValue
+                unit={fiatUnit}
+                value={this.props.totalBalance}
+              />
+            </LText>
+            {data ? (
+              <PriceChange
+                before={startPrice}
+                after={endPrice}
+                color="western"
```
gre: (same)

src/screens/Dashboard.js
```diff
@@ -220,10 +233,12 @@ class Dashboard extends Component<
     const fiatUnit = getFiatUnit("USD");
     const data = getBalanceHistorySum(
       accounts,
-      30,
+      15,
       fiatUnit,
       calculateCounterValue
     );
+    const startPrice: number = data ? data[0].value : 0;
+    const endPrice: number = data ? data[data.length - 1].value : 0;
```
gre: (same)

src/screens/Dashboard.js
```diff
@@ -232,6 +247,11 @@ class Dashboard extends Component<
             <LText semiBold style={styles.balanceTextHeader}>
               <CurrencyUnitValue unit={fiatUnit} value={totalBalance} />
             </LText>
+            <PriceChange
+              before={startPrice}
+              after={endPrice}
+              style={{ color: "white" }}
```
gre: mmh, so this works? 🤔 i thought with the style order above it wouldn't ?

MortalKastor: It works because the `color` prop has a dual use: colorize the value according to the chosen theme, or don't colorize at all if unset. That was unobvious and it's going away.

src/components/PriceChange.js
MortalKastor: I followed the poc mobile screen design, my bad.

PR #33, opened 2018-03-30
wip walletcom-38

NastiaS:


src/index.js
```diff
@@ -49,14 +53,22 @@ export default class Root extends Component<
     console.error(e);
     throw e;
   }
+  startCounterValuePolling = () => {
+    const { store } = this.state;
+    TimerMixin.setInterval(() => {
```
gre: is there a reason why not using `setInterval` directly? https://facebook.github.io/react-native/docs/timers.html#timers doc mention the library TimerMixin, but it's only meant to be used as a react mixin, feature that is now deprecated. So probably we don't need that lib but directly call setInterval.

also one detail but still important, we need to cancel the interval when the app reboot.

basically what you can do is a `clearInterval` at the beginning the reboot call.

app reboot is a feature we use only for one case at the moment: when you go to settings and you logout. we do that because we really want to "reload everything", including the stores. in the future, i think we will do a `"reboot"` when an error occur as well so the app nicely recover from errors,..

src/reducers/counterValues.js
```diff
@@ -18,17 +22,19 @@ const handlers = {
...
 };
 
 const getPairHistory = state => (coinTicker, fiat) => {
   const byDate = get(state, `counterValues.${coinTicker}.${fiat}`);
-  return date => {
-    const d = date || new Date(); // TODO later, we need instead to branch the lookup to pick from latest value
-    return (byDate && byDate[formatCounterValueDay(d)]) || 0;
-  };
+  return date =>
+    (byDate && (date ? byDate[formatCounterValueDay(date)] : byDate.latest)) ||
```
gre: mmh, it does not fallback on `byDate.latest ` in case `byDate[formatCounterValueDay(date)]` is falsy/0. so we need to do things like `date ? byDate[formatCounterValueDay(date)] || byDate.latest : byDate.latest` , a more concise way is `date && byDate[formatCounterValueDay(date)] || byDate.latest`

PR #41, opened 2018-04-11
WALLETCOM-41 Global settings

MortalKastor:
Global app settings for:

- Fiat currency for countervalue (defaulting to the device locale's one, or USD if it's unsupported)
- Time range for the dashboard graph (one week by default)
- Toggle for using red for values going up as used in some part of Asia (off by default, on for 🇨🇳, 🇯🇵, 🇰🇷 and 🇰🇵)


src/reducers/accounts.js
```diff
@@ -71,7 +71,7 @@ export function getAccountById(
 
 export const globalBalanceHistorySelector = createSelector(
   getVisibleAccounts,
-  () => 20, // TODO this will come from settings probably
+  state => state.settings.chartTimeRange,
```
gre: I would create a selector on settings side and import it. exactly like for fiatUnitSelector. idea is that , from this reducers side, you don't really know where things are. this reducer is not aware of other stores structures

src/screens/Settings/ChartTimeRange.js
```diff
@@ -0,0 +1,67 @@
...
+import SectionEntry from "../../components/SectionEntry";
+import LText from "../../components/LText";
+import { saveSettings } from "../../actions/settings";
+import type { SaveSettings } from "../../actions/settings";
+
+const mapDispatchToProps = {
+  saveSettings
+};
+
+const rangeList = {
```
gre: some remarks:
- instead of having to parseInt at some point, probably you should make rangeList an array of `{ name, days }`
- that array should be a prop of the component. and can just set this default in a defaultProps so it's transparent
- that array can be directly passed to the FlatList. you might need to change the keyExtractor to extract the `.days` or something like that. also probably can help a tiny bit to make keyExtractor being a static function & not recreate a reference in render()

src/screens/Settings/SelectFiatUnit.js
```diff
@@ -0,0 +1,52 @@
...
+
+  onPress = (item: *) => {
+    const { navigation, saveSettings } = this.props;
+
+    saveSettings({ counterValue: item.code });
+    navigation.goBack();
+  };
+
+  renderItem = ({ item }) => (
+    <UnitRow unit={item} onPress={() => this.onPress(item)} />
```
gre: this is a global polish, but it would be better if we could render `<UnitRow unit={item} onPress={this.onPress} />` and that inside UnitRow implementation, we would pass `unit` by convention. that way, we don't recreate a function.

this is a general tip to try to use everywhere, especially when using PureComponent-s, but not really important here specifically.

src/screens/Settings/SelectFiatUnit.js
```diff
@@ -0,0 +1,52 @@
...
+  );
+
+  keyExtractor = item => item.code;
+
+  render() {
+    const fiatList = listFiats();
+
+    fiatList.sort((a, b) => a.name.localeCompare(b.name));
```
gre: two problem:

- sort() is mutable. so it will mutate the list returned by listFiats(). if another code somewhere where using that fiatList, it would affect it because I think i don't reclone a new array in its implementation.
- it is done at each time of a render()

so probably you can do `const fiatList = listFiats().slice(0)..sort((a, b) => a.name.localeCompare(b.name))` as a defaultProps , or in a global variable. but not inside render()

src/screens/Settings/index.js
```diff
@@ -0,0 +1,151 @@
...
+      <View style={{ marginVertical: 40 }}>
+        <SectionEntry onPress={this.onSignOut} center>
+          <Text style={styles.signOutText}>Sign Out</Text>
+        </SectionEntry>
+      </View>
+    );
+  }
+}
+
+const SignOut = withReboot(SignOut_);
```
gre: we might want to split a bit the component. like SignOut can be in its own file. i know you didn't modified this, so if you want, we can do it later.

src/screens/Settings/index.js
```diff
@@ -0,0 +1,151 @@
...
+  render() {
+    const { navigation, settings } = this.props;
+    const arrowRight = require("../../images/arrow_right.png");
+
+    return (
+      <ScrollView style={styles.container}>
+        <SectionTitle title="DISPLAY" />
+        <SectionEntry
+          onPress={() =>
+            navigation.navigate({
```
gre: we might want to lighten a bit the code base here. maybe we could have a component per section entry. like `<SelectFiatUnitSection settings={settings} navigation={navigation} />` or something like that? 🤔 

like trying to modularize things a bit more to make it more readable

src/screens/Settings/FiatUnitSection.js
```diff
@@ -0,0 +1,57 @@
...
+import SectionEntry from "../../components/SectionEntry";
+
+const fiatList = listFiats()
+  .reduce(
+    (acc, cur) => [
+      ...acc,
+      { value: cur.code, label: `${cur.name} (${cur.code})` }
+    ],
+    []
+  )
```
meriadec: `.map` should be more suitable for that case
```js
listFiats()
  .map(fiat => ({ value: fiat.code, label: `${cur.name} (${cur.code})` }))
```

src/screens/Settings/FiatUnitSection.js
```diff
@@ -0,0 +1,57 @@
...
+  .sort((a, b) => a.label.localeCompare(b.label));
+
+class FiatUnitSection extends PureComponent<*> {
+  render() {
+    const { navigation, saveSettings, value } = this.props;
+    const arrowRight = require("../../images/arrow_right.png");
```
meriadec: dunno if it's on purpose, but should'nt we require the icon on file scope instead of render?

gre: no, i'm pretty sure this is the correct recommended way. https://facebook.github.io/react-native/docs/image.html
the require() in RN don't behave exactly like node.js i think. i think things gets inlined by the packager.

gre: I mean that it's not a problem to do it in render(), but yeah could also be done with an import, i think both are equivalent in performance

gre: also https://facebook.github.io/react-native/docs/images.html

meriadec: this is the same doc :D

gre: nope :P

gre:
I think it's good to go & we can polish things later

PR #42, opened 2018-04-12
coutervalues refactoring, making app working offline

gre:
- some bugfixes in wallet-common regarding the countervalue api https://github.com/LedgerHQ/ledger-wallet-common/commit/5e7a064c336175aa0f43e351a4100cd62c0cc83c
- refactor countervalues to adopt unified pollRates() (see draft proposal)
- make app working offline, before it was always waiting countervalues to load..
- add a generic CounterValuePollingProvider & withCounterValuePolling
  - code from Root was migrated in the provider. it have the same autopolling feature / listen to appstate like before. however, it only happens when root is "ready", which is when localstorage is asynchronously read (in the future it is important if we decrypt it)
  - counterValuePolling.poll() method is exposed via context, and returns a Promise (we have used it in the "pull to refresh" of dashboard just for PoC)
  - counterValuePolling.polling is a boolean that is true when a pull is happening, we use it to display a spinner on top right header of dashboard (PoC)

Future improvments:
- we should move counterValues actions / stores logic as much as possible into wallet-common, because all wallet-common dependents will benefit from it, also we eventually will migrate to our next countervalue api
- we need to figure out where we want to poll() & where we want to show to the ui it's polling (to be designed)

### DIFF
--- a/src/actions/counterValues.js
+++ b/src/actions/counterValues.js
@@ -1,10 +1,12 @@
 // @flow
+import merge from "lodash/merge";
 import {
   fetchHistodayRates,
   fetchCurrentRates
 } from "@ledgerhq/wallet-common/lib/api/countervalue";
-import { getFiatUnit } from "@ledgerhq/currencies";
 import type { Dispatch } from "redux";
+import { fiatUnitSelector } from "../reducers/settings";
+import type { State } from "../reducers";
 import db from "../db";
 
 export type InitCounterValues = () => (Dispatch<*>) => Promise<void>;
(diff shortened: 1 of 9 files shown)


gre:
don't merge this yet. i need to do more testing with the device tomorrow.

src/actions/counterValues.js
```diff
@@ -22,35 +24,30 @@ export const updateCounterValues: UpdateCounterValues = payload => ({
...
-  const { counterValue } = settings;
-  const currencies = [...new Set(accounts.map(a => a.currency))];
-  fetchCurrentRates(currencies, getFiatUnit(counterValue)).then(data => {
-    dispatch(updateCounterValues(data));
-  });
+  const getLatestDayFetched = cur => {
+    const all = {
+      ...(counterValues[cur.ticker] || {})[fiatUnit.ticker]
+    };
+    delete all.latest;
```
meriadec: consider using [_.omit](http://devdocs.io/lodash~4/index#omit) to avoid creating slow obj

src/components/CounterValuePolling.js
```diff
@@ -0,0 +1,114 @@
...
+import { AppState } from "react-native";
+import hoistNonReactStatic from "hoist-non-react-statics";
+import throttle from "lodash/throttle";
+import { pollRates } from "../actions/counterValues";
+
+// $FlowFixMe
+export const PollingContext = React.createContext(() => {});
+
+export type CounterValuePolling = {
+  poll: () => Promise<*>,
```
meriadec: I'm not sure why `poll` function should be in `state`. It's assigned `this.poll` and is never reassigned

src/actions/counterValues.js
gre: how is it faster? 🤔

src/components/CounterValuePolling.js
gre: yeah, in this component, `this.state` is that value that is provided by context.
if i'm not wrong, in the new context API, you should provide an object that is equal by reference `===` & not creating it each time. `this.state` have this property so it was win-win to use this.state as context value... you can see i did the same for LocaleContext https://github.com/LedgerHQ/ledger-wallet-mobile/blob/master/src/components/LocaleContext.js i'm not sure if it's a pattern found/approved by React team but i'll try to discuss on this subject later with them.

gre: note that CounterValuePolling type, is the type that is exposed by the context, before being a state ;) it could live somewhere else technically, it's just convenient like this.

meriadec: makes sense

src/actions/counterValues.js
meriadec: v8 will less likely optimize property access if some properties are added/deleted after object creation ([more details](https://v8project.blogspot.fr/2017/08/fast-properties.html), [bench](https://jsperf.com/delete-vs-undefined-vs-null/16))

gre: ok interesting. it's temporary code anyway, later we will have the latestDay as part of the meta data that our future api will return (see latest draft) so it will be `O(1)`

PR #67, opened 2018-08-03
Fix staging and release builds on Android

MortalKastor:
Fixed by rolling back Gradle and its Android plugin to the highest version compatible with React Native, and prevent them to try to use a dropped architecture toolchain.

- Android Studio uses the latest NDK version (with no way to force a lower version from the build config files)
- The latest NDK version dropped MIPS support
- Only the latest Gradle Android plugin take that into account
- React Native isn't compatible with the latest Gradle Android plugin
- or even the latest Gradle version

cf android-ndk/ndk#700

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -162,6 +162,15 @@ android {
     dexOptions {
         jumboMode true
     }
+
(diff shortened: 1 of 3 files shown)


meriadec:
Approving but

PR #80, opened 2018-08-13
Settings

NastiaS:
About section. UI as designs.
1) Description row: logo is abstracted out similarly with Desktop 
2) Version row: the version is not dynamic
3) Terms: redirects to Ledger Terms
4) Privacy: redirects to Ledger Privacy
5) Review: will redirect to the App store when there is something to redirect to


src/screens/Settings/About/PrivacyPolicyRow.js
```diff
@@ -0,0 +1,42 @@
...
+import { translate } from "react-i18next";
+import { View, Linking } from "react-native";
+import { createStructuredSelector } from "reselect";
+import SettingsRow from "../../../components/SettingsRow";
+import type { T } from "../../../types/common";
+import colors from "../../../colors";
+import { urls } from "../../../config/urls";
+import ExternalLink from "../../../images/icons/ExternalLink";
+
+const mapStateToProps = createStructuredSelector({});
```
meriadec: we probably don't need it

src/screens/Settings/About/TermsConditionsRow.js
```diff
@@ -0,0 +1,42 @@
...
+import { translate } from "react-i18next";
+import { View, Linking } from "react-native";
+import { createStructuredSelector } from "reselect";
+import SettingsRow from "../../../components/SettingsRow";
+import type { T } from "../../../types/common";
+import colors from "../../../colors";
+import { urls } from "../../../config/urls";
+import ExternalLink from "../../../images/icons/ExternalLink";
+
+const mapStateToProps = createStructuredSelector({});
```
meriadec: same (so no need `connect` at all)

src/screens/Settings/About/index.js
```diff
@@ -2,21 +2,30 @@
...
-import type { NavigationScreenProp } from "react-navigation";
-import SettingsRow from "../../../components/SettingsRow";
+import { ScrollView } from "react-native";
+import AppVersionRow from "./AppVersionRow";
+import LiveReviewRow from "./LiveReviewRow";
+import PrivacyPolicyRow from "./PrivacyPolicyRow";
+import TermsConditionsRow from "./TermsConditionsRow";
+import DescriptionRow from "./DescriptionRow";
 
 const mapStateToProps = createStructuredSelector({});
```
meriadec: same

src/screens/Settings/About/PrivacyPolicyRow.js
NastiaS: yea, you are right. I leave it for some that would probably need it, but the ones that are just for redirection/simple it is not needed

PR #95, opened 2018-08-30
Design react-navigation screens titles

MortalKastor:
- Follows mockup design
- Consistent between ios/android

### DIFF
--- a/src/components/HeaderBackImage.js
+++ b/src/components/HeaderBackImage.js
@@ -1,0 +1,17 @@
+/* @flow */
+import React from "react";
+import { View, Platform } from "react-native";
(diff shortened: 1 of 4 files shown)


src/components/HeaderBackImage.js
```diff
@@ -0,0 +1,17 @@
...
+import React from "react";
+import { View, Platform } from "react-native";
+import ArrowLeft from "../images/icons/ArrowLeft";
+import colors from "../colors";
+
+const HeaderBackImage = () => {
+  const paddingLeft = Platform.OS === "ios" ? 16 : 3;
+
+  return (
+    <View style={{ paddingLeft }}>
```
gre: this could be statically set in a StyleSheet.create

PR #97, opened 2018-08-30
ui for send flow

valpinkman:


src/components/FilteredSearchBar.js
```diff
@@ -0,0 +1,139 @@
...
+
+  state = {
+    focused: false,
+    query: "",
+  };
+
+  constructor(props: *) {
+    super(props);
+
+    this.onChange = throttle(this.onChange, 200);
```
gre: you could inline `onChange = throttle((text:string) => { this.setState(...) }, 200)` down there so you don't need constructor ;)

also, it would be a good practice to have a componentWillMount that will call `this.onChange.cancel()`

src/components/FilteredSearchBar.js
```diff
@@ -0,0 +1,139 @@
...
+  input = React.createRef();
+
+  onFocus = () => this.setState({ focused: true });
+  onBlur = () => this.setState({ focused: false });
+
+  onChange = (text: string) => this.setState({ query: text });
+
+  clear = () => {
+    this.input.current.clear();
+    this.onChange("");
```
gre: polish: you also want to `this.onChange.cancel()`

src/screens/SendFunds/SelectFunds/index.js
```diff
@@ -0,0 +1,218 @@
...
+  static navigationOptions = {
+    title: "Amount",
+  };
+
+  state = {
+    amount: "0",
+    amountBigNumber: new BigNumber(0),
+  };
+
+  isAmountNaN = (amount: string): boolean => isNaN(parseFloat(amount));
```
gre: `isNaN(amount)` would work the same

src/screens/SendFunds/SelectFunds/index.js
```diff
@@ -0,0 +1,218 @@
...
+      amountBigNumber,
+    });
+  };
+
+  render() {
+    const { account } = this.props;
+    const { amount, amountBigNumber } = this.state;
+
+    const isWithinBalance = amountBigNumber.lt(account.balance);
+    const isValid = amountBigNumber.gt(0) && isWithinBalance;
```
gre: Notes for later: we might not be able to do it this way, we will have to ask the "bridge" / the libcore to know if the amount is below balance, because it involve fees, etc..

src/components/FilteredSearchBar.js
valpinkman: fixing this right now

src/components/FilteredSearchBar.js
valpinkman: same here

src/screens/SendFunds/SelectFunds/index.js
valpinkman: fixed

PR #100, opened 2018-08-31
Implement a fallback when no crypto icons

gre:
we can improve the fallback but we probably need one because in future when ERC20 will be there, we won't have icons for all.

capture d ecran 2018-08-31 a 11 56 30

### DIFF
--- a/src/components/CurrencyIcon.js
+++ b/src/components/CurrencyIcon.js
@@ -1,9 +1,10 @@
 // @flow
 
 import React, { PureComponent } from "react";
-import { View } from "react-native";
+import { View, StyleSheet } from "react-native";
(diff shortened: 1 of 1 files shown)


src/components/CurrencyIcon.js
```diff
@@ -15,11 +16,24 @@ export default class CurrencyIcon extends PureComponent<Props> {
...
 }
+
+const styles = StyleSheet.create({
+  altRoot: {
+    alignItems: "center",
+    justifyContent: "center",
+  },
+  altText: {
+    fontSize: 8,
+  },
```
MortalKastor: leftover? 😅

PR #102, opened 2018-08-31
Graph component & PortfolioGraphCard

meriadec:
<p align='center'>
  
</p>

First version of the Graph component for Portfolio (Account page still needs to integrate it). It can receive `isInteractive` prop which make it react to touch events and pass hovered item as callback. It's generated with SVG, backed by d3 utility function to calculate areas/lines.

IMO the Portfolio graph card still needs to be improved:

- no date is displayed when focusing specific point of the graph, only the balance change. cc @bscholving 
- what do in cases where `Infinity` is displayed (totally normal case, because we go from 0 to something, so it's an infinity increase)

Additional components come with the PR:

- `Pills` : horizontal choices list, currently used to choose the time range
- `Delta` : receive a `from` and `to` props, and one of `percent` or `unit`, and display corresponding arrow + corresponding delta
- `Space` : generic space component, can take `h` (height) and `w` (width) props


gre:
I don't understand the second row with percentage and price change when you touch it ? 🤔
I really wonder if we should keep this feature on mobile btw,  it's too much information IMO

meriadec:
I don't get it myself neither, but I believe it's what was asked (I don't see any tooltip in mockups).

But if we remove it, what's the point having an interactive graph :smile: ? We can just have a graph with time range selectors.

gre:
just the graph + the price is useful. the extra derivated info is bonus and can be visually inferred with the graph.

PR #109, opened 2018-09-04
Portfolio animated top bar & Graph improvements

meriadec:
Two features in this commit:

- add scroll-based-animated top bar in Portfolio screen. It displays total balance.
- when touching the graph, the date is shown on the card header

scroll
click

### DIFF
--- a/src/screens/Portfolio/AnimatedTopBar.js
+++ b/src/screens/Portfolio/AnimatedTopBar.js
@@ -1,0 +1,83 @@
+// @flow
+
+import React, { PureComponent } from "react";
+import { StyleSheet, Animated } from "react-native";
+import { translate } from "react-i18next";
+
+import type AnimatedValue from "react-native/Libraries/Animated/src/nodes/AnimatedValue";
+
+import colors from "../../colors";
+import type { T } from "../../types/common";
+
+import LText from "../../components/LText";
+import Space from "../../components/Space";
+import CurrencyUnitValue from "../../components/CurrencyUnitValue";
+
+import type { Summary } from "../../components/provideSummary";
+
+class AnimatedTopBar extends PureComponent<{
+  scrollY: AnimatedValue,
+  summary: Summary,
+  t: T,
+}> {
+  render() {
+    const { scrollY, summary, t } = this.props;
+
+    const opacity = scrollY.interpolate({
+      inputRange: [0, 100],
+      outputRange: [0, 1],
+      extrapolate: "clamp",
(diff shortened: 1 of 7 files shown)


src/screens/Portfolio/index.js
```diff
@@ -86,36 +100,47 @@ class Portfolio extends Component<
...
+              renderItem={this.renderItem}
+              renderSectionHeader={SectionHeader}
+              onEndReached={this.onEndReached}
+              showsVerticalScrollIndicator
+              scrollEventThrottle={16}
+              onScroll={Animated.event(
+                [{ nativeEvent: { contentOffset: { y: scrollY } } }],
+                { useNativeDriver: true },
+              )}
+              ListHeaderComponent={() => (
```
gre: this has to be a "static" component, if not, we're going to have unmount/remount everytime because how React reconcilier works.

I would suggest that the ComputeSummary is not dynamically done in a closure, but instead we do a HOC that send the data by props.

therefore you can have a class field `ListHeaderComponent = () => <GraphCardContainer summary={this.props.summary} />`

src/screens/Portfolio/AnimatedTopBar.js
```diff
@@ -0,0 +1,82 @@
...
+  scrollY: AnimatedValue,
+  summary: Summary,
+  t: T,
+}> {
+  render() {
+    const { scrollY, summary, t } = this.props;
+
+    const opacity = scrollY.interpolate({
+      inputRange: [0, 100],
+      outputRange: [0, 1],
```
gre: you can clamp it because value under 0 or after 1 are useless to update.

src/screens/Portfolio/PortfolioGraphCard.js
```diff
@@ -175,20 +96,29 @@ class PortfolioGraphCardHeader extends PureComponent<{
...
+              value={hoveredItem ? hoveredItem.value : to}
+            />
           </LText>
         </View>
         <View style={styles.subtitleContainer}>
-          <Delta percent from={from} to={to} />
-          <Space w={20} />
-          <Delta from={from} to={to} unit={unit} />
+          {hoveredItem ? (
+            <LText>{moment(hoveredItem.date).format("LL")}</LText>
```
gre: we should make this a pure component that takes a `date`

gre:
smoothy updates! 🍸

meriadec:
too bad that Animated.event can't take spring params :/

PR #115, opened 2018-09-05
Graph on account, and other stuff

meriadec:
<p align='center'>
  
</p>

The graph is slightly different than in portfolio: we display value in currency + countervalue.
This has been opportunity to separate and share GraphCard component, with optional render functions.

Contains also a fix when touching the graph that was interfering with scrolling in the parent View.

Also, add the funny greetings on the portfolio ("good afternoon" etc.).


meriadec:
I've heard that Account page will also need an scroll-based animated header.
Will be part of a future PR.

PR #118, opened 2018-09-06
fix wording mismatch

NastiaS:


src/screens/Portfolio/Greetings.js
```diff
@@ -26,7 +26,7 @@ const getCurrentGreetings = () => {
   } else if (localTimeHour >= eveningBreakpoint) {
     return "common:portfolio.greeting.evening";
   }
-  return "app:dashboard.greeting.morning";
+  return "common:portfolio.greeting.morning";
```
gre: btw i don't think you need to explicitly `common:`

PR #126, opened 2018-09-06
don't show scrollbar in Portfolio

gre:
because makes no sense in infinite scroll

### DIFF
--- a/src/screens/Portfolio/index.js
+++ b/src/screens/Portfolio/index.js
@@ -129,7 +129,7 @@ class Portfolio extends Component<
           renderItem={this.renderItem}
           renderSectionHeader={SectionHeader}
           onEndReached={this.onEndReached}
-          showsVerticalScrollIndicator
+          showsVerticalScrollIndicator={false}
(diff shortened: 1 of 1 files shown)


gre:
to me it is a bug when i see this. on mobile i see no app showing you indicator when the list will grow, it always feel like the app isn't finished

meriadec:
it's the case for many messaging app (WhatsApp, Telegram, etc.) bcause they don't show you the entire message list unless you scroll (seems that is exact same behaviour as txs list), and when you scroll, the scrollbar thumb size changes.

even google mobile (with "Show more") works like that (I mean: displaying scrollbar)

IMO it's accessibility stuff, not a fan of hiding this subtle info (we do it on desktop and this is bad)

MortalKastor:
I can't think of any app hiding the scrollbar on infinite load 🤔
From what I see, popular apps based on a huge infinite list (I checked Twitter, Instagram and Tumblr) show the scrollbar.

gre:
ok, we can show it back i guess then!

PR #130, opened 2018-09-06
new help settings design and refactor

NastiaS:
1) Re-implement Help according to new design
2) Refactoring Icon in the Cirlular Container in the app


src/components/IconInCircularCont.js
```diff
@@ -0,0 +1,35 @@
+/* @flow */
+import React, { Component } from "react";
+import { StyleSheet, View } from "react-native";
+
+type Props = {
+  icon: any,
+  bgIconContainer: string,
+  size: number,
+};
+class IconInCircularCont extends Component<Props> {
```
gre: we can probably call it `Circle` and rename icon props to `children` to write code like `<Circle size..><Icon />`
also maybe rename `bgIconContainer` to `bg`

PR #134, opened 2018-09-07
various fixes: keyboardavoidingview, blur text input, asc|desc UI rev…

valpinkman:
…ersed, currency icons in accountlist

### DIFF
--- a/src/components/OperationRow.js
+++ b/src/components/OperationRow.js
@@ -1,13 +1,15 @@
 /* @flow */
 import React, { PureComponent } from "react";
 import { View, StyleSheet, TouchableHighlight } from "react-native";
+import { getOperationAmountNumber } from "@ledgerhq/live-common/lib/helpers/operation";
 
 import type { Account, Operation } from "@ledgerhq/live-common/lib/types";
 
 import LText from "./LText";
-import OperationIcon from "./OperationIcon";
 import CurrencyUnitValue from "./CurrencyUnitValue";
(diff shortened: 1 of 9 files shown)


src/components/OperationRow.js
```diff
@@ -56,6 +56,7 @@ class OperationRow extends PureComponent<Props, *> {
               ellipsizeMode="tail"
               style={[styles.topLine, { color: valueColor }]}
             >
+              {isIn ? "+ " : "- "}
```
gre: This shouldn't be done like this, in fact it's part of formatCurrencyUnit to do this (there is utilities in live-common to have the actual value from the operation + a special mode to force display of "+")

Idea of this is that potentially the way this is formatted depends on the locale

src/screens/SelectFunds/index.js
```diff
@@ -90,67 +98,74 @@ class SelectFunds extends Component<Props, State> {
     const isWithinBalance = amountBigNumber.lt(account.balance);
     const isValid = amountBigNumber.gt(0) && isWithinBalance;
 
+    const keyboardVerticalOffset = Platform.OS === "ios" ? 60 : 0;
```
gre: Interesting. Any idea why 60 ?

valpinkman: because 40 wasn't enough

src/components/OperationRow.js
valpinkman: ok I need to check this out then

PR #143, opened 2018-09-12
bump react native and react

valpinkman:


package.json
```diff
@@ -51,9 +51,9 @@
     "numeral": "2.0.6",
     "prando": "^3.0.3",
     "querystring": "^0.2.0",
-    "react": "16.3.2",
+    "react": "16.5.0",
     "react-i18next": "7.10.1",
-    "react-native": "0.55.4",
+    "react-native": "0.57-RC",
```
gre: how does this work? latest is `0.57.0-rc.4` , is it a semver thing? 🤔

src/screens/SelectRecipient/index.js
```diff
@@ -95,7 +96,9 @@ class SelectRecipient extends Component<Props, State> {
                 console.log("scan qr code"); // eslint-disable-line no-console
               }}
             >
-              <Icon name="qrcode" size={16} color={colors.live} />
+              <Text>
+                <Icon name="qrcode" size={16} color={colors.live} />
+              </Text>
```
gre: any reason to put that in a Text?

valpinkman: Yes, it crashes because Icon is a font base icon and uses text to render, so it breaks iOS otherwise (didn't test on android without the <Text>)

package.json
valpinkman: will fix but I do have the correct one installed

valpinkman:
react native just released 0.57.0 so no need to use RC anymore

gre:
this PR will wait next week. it is overall conclusive but we'll need to polish a bit.

- some bugs with libcore to focus on in develop
- make Jest working again
- intensively test the app and look for regressions (iOS and Android)
- check the Android build and if there is no regression & potentially update some Android build tool (@MortalKastor have great experience in it)

valpinkman:
> * make Jest working again

fixed with https://github.com/LedgerHQ/ledger-live-mobile/pull/143/commits/736f49a3fee943b3b7643b1bfc06dafb7bb1f67e

PR #148, opened 2018-09-12
Libcore integration for Android & iOS

meriadec:
- new dependency on `lib-ledger-core-react-native-bindings`
- load the dynamic library at runtime
- get the `core` cross-platform singleton from `logic/getLibcore.js`, which init the wallet pool instance and a bunch of implementations (http handler, logger..) and exposes our native methods
- add a basic bitcoin account sync from xpub POC in `helpers/libcore.js`
- add a little hack to prevent (showing) error from double promise callback call on libcore side (see `src/polyfill.js`)

I have only tested it on Android, the iOS native methods still needs to be mapped in `logic/getLibcore.js`, if I remember the generated names are different so @KhalilBellakrid  we'll need you to point us the direction where to search :)

Anyway, the whole thing should work out of the box with only a `yarn install`.

### DIFF
--- a/src/logic/getLibcore.js
+++ b/src/logic/getLibcore.js
@@ -1,0 +1,72 @@
+import { Platform, NativeModules } from "react-native";
+
+let core;
+let walletPoolInstance;
+let threadDispatcher;
+
+export default async () => {
+  if (core) {
+    return core;
+  }
+
+  if (Platform.OS === "android") {
+    core = {
+      getPoolInstance: () => walletPoolInstance,
+      getThreadDispatcher: () => threadDispatcher,
+      flush: async () => {
+        await core.coreHttpClient.flush();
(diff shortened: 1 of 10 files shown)


src/logic/getLibcore.js
```diff
@@ -0,0 +1,72 @@
...
+let core;
+let walletPoolInstance;
+let threadDispatcher;
+
+export default async () => {
+  if (core) {
+    return core;
+  }
+
+  if (Platform.OS === "android") {
```
gre: (future notes. we can probably have split into two files getLibcore.ios.js and getLibcore.android.js )

meriadec: I thought of it, but as there is a bit of common logic (instanciating wallet pool & pal) I chose to go like that

PR #638, opened 2018-12-14
Tentative react-navigation upgrade, with issues

juan-cortes:
## Update
image

Following [the guidelines for an authentication flow](https://reactnavigation.org/docs/en/auth-flow.html) I've refactored the logic (the flags checking) into its own component and made the call of what screen to choose over there. This was needed because with the latest version of the libraries we were triggering an instant crash in Android/iOS? on launch, when it tried to navigate to onboarding from portfolio.

------------------

### What's done
Bumping `react-navigation` to v3, building on what @gre did over at https://github.com/LedgerHQ/ledger-live-mobile/pull/541 . There were some things that had to be done differently though. This update allows us to remove the temporary android screen transitions since they no longer cause that weird glitch with the components while animating (worth checking on real devices affected by this before).

It also brings the much-wanted feature tap top scroll to top on the whole app, basically, any screen containing a scrollable view and a header will have the feature of scrolling to the top when touching the header, as it can be seen in the demo.

I've gone through the whole app and I've fixed everything that seemed out of order, the app seems more responsive on transitions, but that might just be me thinking it's faster, I haven't compared them.

~## What is not working
Here's the kicker.
For whatever reason https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-navigation_v3.x.x/flow_v0.60.x-/react-navigation_v3.x.x.js is not up to date, nor is the source for that back in https://github.com/react-navigation/react-navigation/blob/master/flow/react-navigation.js . Either they're not up to date or they don't plan to include the native components that are wrapped to have this scroll to top feature added.~

This means there are a gazillion `// $FlowFixMe` on all the imports of scrollables from `react-navigation`. 

### How to test it?
Apart from running `yarn` you'll need to upgrade the type definitions by running `yarn sync-flowtyped`. After that you might (I had to) need to delete your `node_modules` folder completely if it's acting up. If it starts, then we're good.

### Demo
v3


juan-cortes:
@gre Rebased with develop to update this pr, it's a bit of a pain because this is on the old fork so it's another remote to maintain on my local. More chances to break things unknowingly. I will test it tomorrow morning again but it should be fine.

gre:
Cool thanks! Plan is to merge this after we do a build.

gre:
@Arnaud97234 this should be good to gets in.

PR #662, opened 2018-12-28
Developer testing/debug tool for SVG icons from the app

juan-cortes:
image

## What's done
- New screen
- New menu item
- Generator script that updates the screen
- Refactored all Icons I could find and gave them default values to allow them to be called without the props, most of them are always of the same size.
- A first round of renaming, there were icons with completely wrong names inside, this didn't affect us since we rename the default export at import but it makes more sense to be consistent.
- Make the script run from a yarn command so we don't have to invoke it manually
- Make the screen use a grid format instead of rows so you get more bang for your buck
- A highlight for icons inside `/icons` using bitmap graphics

## What's not done
- Taking into account bitmap images that we are using, this screen could alert us of non-vector images in case we want to update them. Some of the logos shown are wrapped pngs/jpegs inside a component, especially for logos.
- Make sure all the Icons are located inside `Icons` and not somewhere else, I did my best but I think I might have left one or two around.
- For non-square icons, some of them still use a single `size` prop, meaning the will look distorted when rendered, we need to take a look at them.
- Perhaps implement a search mode
- Address inconsistent naming
- Find incorrect rendering of some icons, I've fixed a few incorrect props on the Svgs that failed to convert the name from dash to camelcase resulting in wrong widths/colors/etc.

## How Does it stay updated
After we create/rename an icon we need to run `yarn debugSVG` to regenerate the `DebugSVG` screen that lists all the icons alphabetically. ~Make sure to run prettier afterward since I can't predict the line breaks for some of the imports.~ The output is piped to prettier first now so it won't break the CI.


gre:
@juan-cortes  do you want to keep this branch up to date so we can merge it? i think it's legit usecase :)

juan-cortes:
I was looking into it the other day, since we changed so much, I think I'm killing this pr and launching one from scratch, bringing only the generator and such

PR #715, opened 2019-01-28
Added the `ci` command for yarn, with the added integrity check

juan-cortes:
With this, we can call `yarn ci` and run prettier, lint, flow and the integrity check for dependencies in case we haven't run `yarn` in a while to avoid pushing an outdated lock, left the `test` reference regardless of it being a placeholder for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}
 
       # run tests!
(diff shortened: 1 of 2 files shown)


gre:
can you replace the `yarn ci` command to be run from https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/.circleci/config.yml ? it's how circleci know what command to run

juan-cortes:
Oh sure, I can do that

PR #718, opened 2019-01-29
LL-942 Filtered out empty accounts

juan-cortes:
To be consistent with what happens if you go to a specific account with no funds (only receive action is available), we should filter the accounts displayed on the select account screen for send flow.


gre:
good catch

PR #720, opened 2019-01-29
Import desktop accounts: align wording

dasilvarosa:
- Update to align mobile app with desktop settings. 
- Introducing **LiveQR code**

**Dev To Do**
- [ ] remove `import.descBottom`

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -432,8 +432,8 @@
       "cta": "Open Ledger Live"
     },
     "stepScanQR": {
-      "title": "Please go to <1>Settings > Export accounts</1> in the desktop app and scan the QR code",
-      "cta": "Scan QR code"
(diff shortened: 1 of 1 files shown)


src/locales/en/common.json
```diff
@@ -574,12 +574,11 @@
...
       "scan": {
-        "title": "Scan QR Code",
+        "title": "Scan LiveQR code",
         "descTop": {
-          "line1": "Please open Ledger Live desktop and go to",
-          "line2": "Settings > Mobile > Export"
+          "line1": "In Ledger Live desktop, go to:",
+          "line2": "Settings > Export accounts > Export"
         },
-        "descBottom": "Please put the QR code within the square"
```
gre: this is a removal of wording. it affects the design. we need a second eye from design and then we need a dev to remove the location if approved.

dasilvarosa: Design is already OK : https://zpl.io/2jw9wqO

PR #721, opened 2019-01-29
LL-842 receive flow - share address

valpinkman:
### Type

Feat

### Context

LL-842

Receive Flow

### DIFF
--- a/src/components/ShareLink.js
+++ b/src/components/ShareLink.js
@@ -1,0 +1,54 @@
+// @flow
+
+import React, { PureComponent } from "react";
+import { StyleSheet, Share } from "react-native";
+import Icon from "react-native-vector-icons/dist/Feather";
+import Touchable from "./Touchable";
+import LText from "./LText";
+import colors from "../colors";
+
+type Props = {
+  style?: *,
+  children: string | React$Element<*>,
+  string: string, // String to be shared
+};
+
+class ShareLink extends PureComponent<Props> {
+  onPress = async () => {
+    const { string } = this.props;
+    await Share.share({
+      message: string,
+    });
+  };
+
+  render() {
+    const { style, children } = this.props;
+
+    return (
+      <Touchable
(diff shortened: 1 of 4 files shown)


src/components/ShareLink.js
```diff
@@ -0,0 +1,54 @@
...
+import React, { PureComponent } from "react";
+import { StyleSheet, Share } from "react-native";
+import Icon from "react-native-vector-icons/dist/Feather";
+import Touchable from "./Touchable";
+import LText from "./LText";
+import colors from "../colors";
+
+type Props = {
+  style?: *,
+  children: string | React$Element<*>,
```
gre: React$Node

src/components/ShareLink.js
```diff
@@ -0,0 +1,54 @@
...
+  string: string, // String to be shared
```
gre: chose a better name than `string` for the prop, maybe `value`

src/components/ShareLink.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+  };
+
+  render() {
+    const { style, children } = this.props;
+
+    return (
+      <Touchable
+        event="ShareLink"
+        style={[styles.linkContainer, style]}
```
gre: remove the `style` prop. it's not needed from the user side so no need to add this complexity

can just `style={styles.linkContainer}`

src/components/ShareLink.js
```diff
@@ -8,26 +8,25 @@ import LText from "./LText";
 import colors from "../colors";
 
 type Props = {
-  style?: *,
-  children: string | React$Element<*>,
-  string: string, // String to be shared
+  children: string | React$Node,
```
gre: React$Node includes `string` in it so you can just set the whole type as `React$Node`

valpinkman: shit... will change

PR #726, opened 2019-01-30
LL-919 Upgrade important libraries

juan-cortes:
Updated:

- `"jsc-android"`: "^236355.1.1",
- `"react-native-camera"`: "1.9.2",
- `"react-native-gesture-handler"`: "^1.0.15",
- `"react-native-svg"`: "9.0.6",

There are a few errors and crashes that I'm currently going through:
- [x] Camera requires to specify that we don't want audio
- [x] Crash after importing the LiveQR Code, for some reason `Svg` was complaining about passing a spread object from props, it didn't seem to be used anyway, removing it fixed it. This comes from ` StyleSheet.create` not returning a plain js object, breaking the spread operator. It noted that the error was a performance optimization and not spec compliant. 🤷‍♂️ 
- [ ] Occasional weird crash relating to styles mostly that doesn't happen always but means we can't ship this.


juan-cortes:
@gre no idea why flow complains about the camera library, wdyt?

gre:
try to refresh the flow def

yarn sync-flowtyped

juan-cortes:
> yarn sync-flowtyped
If only it were that easy... The only way I can see of getting rid of this is ignoring the whole library in `.flowconfig`, but if we can avoid it..

gre:
I think ignore the lib def for now.
The latest version is recent, let's give some time they fix the types

gre:
`StyleSheet.create` should not be assumed to return a `{}`, i'm pretty sure it used to return numbers (the doc says it returns IDs https://facebook.github.io/react-native/docs/stylesheet#create ), I've seen a few place where we exploded styles like the one coming from getTextStyle, we'll have bad surprise.

only way to combine styles should be by using `[style1, style2]` and not `{...style1, ...style2}`, I wish there were a eslint to assess that, i'm pretty sure we have that at bunch of places.

gre:
we were not able to reproduce the crash, we will fix if we see again later. good to merge

PR #713, opened 2019-01-28
LL-922 fix clear input in search bar

valpinkman:

PR #729, opened 2019-01-30
LL-922

juan-cortes:
I can no longer push to val's branch, probably something to do with the recent reforking we had to do 🤷‍♂️. I get a permission denied yada yada. In any case, this adds a call to the input cleared event of our custom implementation of the text input.

Fixes https://github.com/LedgerHQ/ledger-live-mobile/pull/713


gre:
`onInputCleared` , well, we missed that one @valpinkman

PR #739, opened 2019-02-04
Fixed dead Play store URL

Luuk2016:
### Type

Dead URL

readme :trollface:


juan-cortes:
Nice find, we had that ancient url in the codebase for a while, it was also in the "rate this app" cta somewhere. 🤦‍♂️

PR #741, opened 2019-02-04
Bump react-navigation

juan-cortes:
It should fix all the issues from their [CHANGELOG](https://github.com/react-navigation/react-navigation/blob/master/CHANGELOG.md), I also tried re-enabling the `useScreens` thingy but it's still breaking text selection, so that's still there.


gre:
@juan-cortes for the conflict you can attempt to just run `yarn`

PR #701, opened 2019-01-15
Fixes LL-655

gre:
flush the (un)install when interrupted

PR #742, opened 2019-02-05
LL 655

gre:
#701


gre:
damn. why does it work for me.

gre:
i've added a delay of 1s to see if it helps in any way.

PR #745, opened 2019-02-06
LL-993 Added intent-filter to allow handling ledger devices via USB OTG

juan-cortes:
image

### Only for Android
On the current develop version of the app, the experimental usb connections request user permission to access the connected device *everytime we want to interact with it*, meaning that if we disconnect the device and go to the manager again it will once again request user confirmation to access the device. The checkbox provided on that dialog seems to be ignored.

By setting an `intent-filter` on our activity that listens to the `USB_DEVICE_ATTACHED` event for a list of vendorId/productId pairs we can request the user to set Ledger Live Mobile as the default app when plugging in (and unlocking) the device, allowing us to skip the permission request when trying to interact with it.

It's a quick win that makes the ux better.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -42,6 +42,11 @@
             <category android:name="android.intent.category.BROWSABLE"/>
             <data android:scheme="ledgerhq"/>
         </intent-filter>
+        <intent-filter>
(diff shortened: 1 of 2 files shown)


juan-cortes:
I'll look into it, it has to do with how Android handles that even by relaunching the activity, I'll see what I can do. Might update this pr with the hid non blocking part too since it's related

PR #751, opened 2019-02-07
Make the pairing screen only strictly for BLE transport

gre:
the pairing flow should only be for the ble devices (not the other transports like http proxy, usb, etc.)

### Type

bug fix

that BLE pairing still works, that you can also still see "already paired" device in the list.


gre:
@Arnaud97234 I'm not understanding the report. it's only on "scanning bluetooth devices" list that is affected, Manager list should still display http proxy devices. they just should not appear in Pair a new device screen

PR #752, opened 2019-02-08
Upgrade react-native-hid implementation

gre:
non blocking implementation of HID.

### Type

improvement

### Context

https://github.com/LedgerHQ/ledgerjs/pull/288

- test that the usb implementation still works (install apps, receive,..)
- test that you can effectively tap on the "Copy" address during a Receive and when you have not yet accepted on the device. (before it was blocked, the ui was actually stucked)


juan-cortes:
Testing it directly without applying https://github.com/LedgerHQ/ledger-live-mobile/pull/753 I get random crashes, but the USB connectivity seems to work fine. There are some cases (connect the device on manager) where if you cancel mid verification it doesn't seem to want to connect again until you unplug/plug the device again.

Perhaps consider a future improvement of detecting the canceled exchange and showing some UI warning indicating that they might need to replug the device.

PR #753, opened 2019-02-08
Bumped versions of react, react-native, react-native-svg, react-test-renderer

juan-cortes:
This seems to solve the random crashes we were getting from svg and such.
Everything seems to work perfectly still


gre:
we'll let Arnaud test this. After that PR merged, i'll do another PR to bump ledgerjs with https://github.com/LedgerHQ/ledgerjs/pull/284 that contains important fixes for Bluetooth

Issue #748, opened 2019-02-06
rename: yarn ci => yarn test


gre:

meriadec:
but that will override our awesome [test suite](https://github.com/LedgerHQ/ledger-live-mobile/blob/9e8a655860f4d71940db206e33b92ff1d232b379/package.json#L23) lol

PR #758, opened 2019-02-08
rename: yarn ci => yarn test

juan-cortes:
@gre asked for this. Fixes https://github.com/LedgerHQ/ledger-live-mobile/issues/748


juan-cortes:
`circleci` fails because it wants to run `yarn ci` and it no longer can

PR #760, opened 2019-02-09
LL-968 Allow user to finish onboarding with an empty import

juan-cortes:
This ads the posibility to retry after scanning the LiveQR Code from desktop or to import simply the settings in the case of not having accounts to import. It initially didn't have the retry cta, hence the comments below mentioning the need to reword.

image

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -33,7 +33,8 @@
     "upToDate": "Up to date",
     "outdated": "Outdated",
     "satPerByte": "sat/bytes",
-    "notAvailable": "Not available"
+    "notAvailable": "Not available",
+    "import": "Import"
   },
   "errors": {
     "AccountNameRequired": {
@@ -594,7 +595,7 @@
         "descBottom": "Please put the LiveQR code within the square"
       },
       "result": {
-        "title": "Select accounts",
+        "title": "Import accounts",
         "newAccounts": "New accounts",
         "updatedAccounts": "Updated accounts",
         "empty": "Add a new account",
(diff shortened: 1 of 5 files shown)


khalilbenihoud:
We should reword it since we don't really give the users the possibility to try again. @aramab

aramab:
@khalilbenihoud 
See link in the design issue for wording and retry button : https://ledgerhq.atlassian.net/browse/LL-969?focusedCommentId=23317&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23317

juan-cortes:
Do we keep this one on hold until @khalilbenihoud finishes the sandbox?

khalilbenihoud:
@juan-cortes  - @bscholving is going to take care of it.

juan-cortes:
great!

juan-cortes:
@gre @Arnaud97234 @khalilbenihoud This is ready for testing now, I've updated the first message with the new screenshot

juan-cortes:
@Arnaud97234 nice catch, I'm guessing it _should_ be exportable, we could make it exportable, but do we care enough to do it? @gre

gre:
Yes, let's export Developer Mode settings as well, it makes sense because it is somewhat consistent with account you imports. If you import only Testnet account, then later you go to Manager, you don't see Testnet app.

so let's import it, however @Arnaud97234  please don't consider it blocking this PR. Because to actually fix this, we first need to provide a fix in desktop project (it doesn't export this info) and then we can also do a new PR on mobile.

the fix is also to be done in live-common btw @juan-cortes  , look at `src/cross.js`

PR #764, opened 2019-02-12
New version of LedgerJS transports

gre:

PR #766, opened 2019-02-12
fix account name length + ui fixes

valpinkman:
Augment the max number of character for the account name to 50 (to be iso with desktop) and fixed SettingsRow to accept longer name

### Type

feat + fix

### Context

LL-951 LL-952

EditAccountName, Accounts page


gre:
( hold because we need to focus on #764 )

Arnaud97234:
It miss the ellipsis ?

gre:
I guess there are many place we need to make sure the UI correctly ellipsis. Cc @valpinkman

PR #773, opened 2019-02-15
Fixed a regression in SettingsRow

juan-cortes:
A past fix for Account name length brought in a regression into all settings row (Settings/General for example). The solution is to not center the elements of the row inside but rather space them, and a few more tweaks. This limits the right content to a max of 50% too, so it never squeezes the left either.

image

### DIFF
--- a/src/components/SettingsRow.js
+++ b/src/components/SettingsRow.js
@@ -120,6 +120,7 @@ const styles = StyleSheet.create({
     paddingVertical: 16,
   },
   textBlock: {
+    marginRight: "auto",
     paddingRight: 16,
     flexShrink: 1,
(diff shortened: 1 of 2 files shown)


juan-cortes:
Oh damn, that looks great to me :trollface:

juan-cortes:
I didn't account for rows with three elements, so the `justify-content` for space will indeed center the middle one. The solution here is a bit obscure, adding a `marginRight` auto to the middle element, makes it go to the left. Don't flex me why.

@Arnaud97234 @valpinkman 
image

gre:
ok i'll assume this done and we can test more on develop

PR #774, opened 2019-02-15
LL-1028 Added block to 0 fees

juan-cortes:
Disabled the button and show an error when 0 fees are entered in the custom fees for bitcoin family. **Updated** with the latest design from @bscholving which always shows the warning and will turn red on invalid input, updated wording too.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     }
   },
   "dependencies": {
-    "@ledgerhq/errors": "4.38.0",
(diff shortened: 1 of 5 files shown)


juan-cortes:
Pinged @bscholving for a UI solution to this

juan-cortes:
@gre @Arnaud97234 should be done now with latest design

PR #775, opened 2019-02-15
Autolock loop - Simplified the autolock logic, removed the timeout

juan-cortes:
The app state cycle was a bit messy and the `setTimeout` wasn't triggering consistently. Since it wasn't working at all on iOS and we had a little delay when coming back until it was locked (that was deemed acceptable at the time) and since Android handles the events differently and locks a lot faster I simplified the logic a lot, removing the timeout and handling the app state changes tighter.

This should fix the lock loop that @aramab was experiencing and also make the autolock work more consistently in both platforms.

image

### DIFF
--- a/src/context/AuthPass/index.js
+++ b/src/context/AuthPass/index.js
@@ -18,6 +18,7 @@ const mapStateToProps = createStructuredSelector({
 type State = {
   isLocked: boolean,
   biometricsError: ?Error,
+  appState: string,
 };
 
 type Props = {
@@ -34,6 +35,7 @@ class AuthPass extends PureComponent<Props, State> {
   state = {
     isLocked: !!this.props.privacy && !wasUnlocked,
     biometricsError: null,
+    appState: AppState.currentState || "",
   };
 
   static getDerivedStateFromProps({ privacy }, { isLocked }) {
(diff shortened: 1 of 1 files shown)


gre:
@juan-cortes can you squash into one commit so it's easier to cherry-pick? i'll cherry-pick from master on monday so we do the patch release

juan-cortes:
@arnaud97234 doubt I can do anything about that, @gre I will squash later, sure

gre:
It's an interesting feedback but I don't feel it's a regression right?
Nevertheless, let's add this in our backlog.

One idea to fix this is to delay the splashscreen.show() , even tho I'm not even sure if it's used at this time but that's my first blind guess

juan-cortes:
I'll take a look tonight, maybe I can do something about it, but I think it's just the time it takes for the unlock to do its thing. I'll give it a couple hours see what comes out of it

juan-cortes:
I'll take a look tonight, maybe I can do something about it, but I think it's just the time it takes for the unlock to do its thing. I'll give it a couple hours see what comes out of it

juan-cortes:
It might be possible to use the app state changes to animate an overlay that hides that screen if it's an issue or the sensitive data that could be seen in that fraction of time that it takes to lock, I tested a bit the idea and it should be doable. I say **keep it as it is**, and see what kind of feedback from the community we get. See if it's an issue at all.

@gre squashed into a cherry-pickable commit

gre:
You don't need to implement an overlay btw. It's already what splashscreen.show() and hide () does

And btw we already have a timeout of 300ms today. Maybe you want to try increasing that. See code of src/index

juan-cortes:
afaik the splashscreen is only used from the `onRebootStart` from js and natively called at launch, so unless we are doing a resetAll or clear data, the splash wouldn't show? Maybe I'm misunderstanding it, I was playing with the idea of showing something that fades into transparency on unlock, but it's probably not even worth it

gre:
In Arnaud gif we see the splash screen appearing no? 🤔

juan-cortes:
Then I'm all confused again! I don't get the splash on an autolock -.-

gre:
I suspect it's only happening if you come back after a long term (app reboot)
So yeah, not tricky to solve , let's leave that in a backlog task as initially said ;)

PR #781, opened 2019-02-19
LL-1046 Override detected language if it's RTL

juan-cortes:
### Issue
On devices (Android) using a right-to-left language such as Arabic or Hebrew, Android enforces a conversion of the layout that breaks the app making it look terrible. 

### Proposed solution
Common solutions proposed online to disable the rtl conversion seem to not currently work, as a workaround we will detect the locale of the device `onResume` at the native level and enforce the use of `en` instead for rtl languages. iOS seems to not be affected by this and respects the original app's layout instead of enforcing a conversion (kudos apple)

### Caveats
I left a `TODO` in the code for this, but if we ever support a rtl language we'd have to add the exception to the code in order for the translations to actually work, otherwise it will use this and end up with english. 

### Demo
image

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -23,6 +23,7 @@
       android:name=".MainApplication"
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
+      android:supportsRtl="false" tools:replace="android:supportsRtl"
       android:allowBackup="true"
       android:theme="@style/AppTheme">
       <activity
--- a/android/app/src/main/java/com/ledger/live/MainActivity.java
+++ b/android/app/src/main/java/com/ledger/live/MainActivity.java
@@ -3,15 +3,20 @@ package com.ledger.live;
 import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
+import android.content.res.Configuration;
 import android.os.Bundle;
+import android.view.View;
 import android.view.WindowManager;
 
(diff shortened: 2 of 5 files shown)


juan-cortes:
I think the testing should only consider starting the app from scratch not switching while it's open so some of those glitches should be disregarded. As for the rest, I can get a hold of a Xiaomi and test tomorrow morning, see why it's not pretty still

gre:
@Arnaud97234 let's not support the "hot swap" of translation if it's ok. only test "opening the app with Hebrew" case

juan-cortes:
The operation details date is because we are using `toLocaleDateString`, I would suggest considering this a minor drawback of not having the app completely broken and merge this already :trollface: @gre

gre:
Agreed. Also it's a bit weird toLocaleDateString is enabled even for languages we don't support. We will need to find a way to fallback but we'll see later

gre:
@Arnaud97234 ok to merge this if we ignore the Date bug? we'll add a task in our backlog that is to make sure the Date formatter don't format to unsupported langs

gre:
@juan-cortes actually can you try to pass "locales" option in toLocaleDateString ? It seems it's possible to provide it like this.

We would pass the language we have from the store, that way we would fallback to English I think.

juan-cortes:
@gre Your idea worked, passing the list of supported locale ids makes the formatted date adhere to one of them. Nice!
@Arnaud97234 can you take a look and see if operation details look ok now? Thank you

PR #783, opened 2019-02-19
LL-1051 Created SkipLock component to disable auto-lock in certain processes

juan-cortes:
Created a `<SkipLock/>` component that will allow us to disable the autolock behaviour in certain parts of the application to avoid breaking a flow or disturbing a process that requires the app to be inactive (pairing for example could also get this if wanted).

To use it simply render the component in the screen you want to disable autolock for.

### How to test?
Autolock should be disabled while installing an app, updating firmware, or validating data on the device. To test, simply navigate out of the app while on one of those screens, wait 1 minute (or modify the environmental variable to a lower setting) and come back to the app.

### DIFF
--- a/src/components/behaviour/SkipLock.js
+++ b/src/components/behaviour/SkipLock.js
@@ -1,0 +1,52 @@
+// @flow
+import React, { PureComponent } from "react";
+import { withNavigationFocus } from "react-navigation";
+
+export const SkipLockContext = React.createContext({
+  skipLockCount: 0,
+  setEnabled: (_: boolean) => {},
+});
+
+class SkipLock extends PureComponent<*> {
+  render() {
+    return (
+      <SkipLockContext.Consumer>
+        {context => (
+          <SkipLockInner setEnabled={context.setEnabled} {...this.props} />
+        )}
+      </SkipLockContext.Consumer>
+    );
+  }
+}
+
+class SkipLockInner extends PureComponent<{
+  setEnabled: (enabled: boolean) => void,
+  isFocused: boolean,
+}> {
+  lastValue = false;
+
+  componentDidMount() {
+    this.report(true);
+  }
+
+  componentDidUpdate() {
+    this.report(this.props.isFocused);
+  }
+
+  componentWillUnmount() {
+    this.report(false);
+  }
+
+  report = enabled => {
+    if (this.lastValue !== enabled) {
+      this.props.setEnabled(enabled);
+      this.lastValue = enabled;
+    }
+  };
+
+  render() {
+    return null;
+  }
+}
+
+export default withNavigationFocus(SkipLock);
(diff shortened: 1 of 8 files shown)


src/context/AuthPass/index.js
```diff
@@ -123,7 +135,12 @@ class AuthPass extends PureComponent<Props, State> {
         />
       );
     }
-    return children;
+    const contextValue = { skipLockCount, setEnabled };
```
gre: I would just pass in setEnabled in the context to avoid rerendering all consumers (because new object ref)

src/components/behaviour/SkipLock.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+class SkipLockInner extends PureComponent<{
+  setEnabled: (enabled: boolean) => void,
+  isFocused: boolean,
+}> {
+  lastValue = false;
+
+  componentDidMount() {
+    this.report(true);
```
gre: It should be this.props.isFocused

src/context/AuthPass/index.js
juan-cortes: RIght, gotcha

juan-cortes:
@gre done

juan-cortes:
@Arnaud97234 don't take that bug into consideration for testing this pr, this pr is only about disabling unlock for certain screens, it's currently implemented (easiest to test) on the Debug->Mock screen, leaving it open on that screen should never autolock.

juan-cortes:
I've been trying to repro this bug on my end, it's not happening at all. @gre wdyt?

gre:
So I think the bug might be here:

https://github.com/LedgerHQ/ledger-live-mobile/blob/43cb65f4743e2e1cf7d1011c45c4c1d10125161d/src/context/AuthPass/index.js#L66-L78

I think in the case Arnaud reproduce, the (iOS) app will only trigger `"inactive"`, so we need to record this too.
It sounds like this `appInBg` should always be set each time we enter this `handleAppStateChange` (whatever is the app state changes)

so maybe this works?

```js
 handleAppStateChange = nextAppState => { 
   if ( 
     this.state.appState.match(/inactive|background/) && 
     nextAppState === "active" && 
     this.appInBg + AUTOLOCK_TIMEOUT < Date.now() 
   ) { 
     this.lock(); 
   }
   this.appInBg = Date.now(); 
   this.setState({ appState: nextAppState }); 
 };
```

juan-cortes:
Resetting the timer for `appInBg` on inactive will reset it when coming back from background to active too (active->inactive->background but also background->inactive->active). Which would, in turn, disable the timer.  We can check the previous state too but don't think it would guarantee it either.

src/constants.js
```diff
@@ -17,7 +17,7 @@ export const GENUINE_CHECK_TIMEOUT = 120 * 1000;
 export const GET_CALLS_RETRY = 3;
 export const GET_CALLS_TIMEOUT = 60000;
 
-export const AUTOLOCK_TIMEOUT = 60000;
+export const AUTOLOCK_TIMEOUT = 6000;
```
gre: oops

juan-cortes: Oh damn -.- it happened _again_

PR #784, opened 2019-02-20
LL-983 Sentry reports, link cleanups

juan-cortes:
Added more reported errors to Sentry, removed unneeded catch blocks for links.

### DIFF
--- a/src/libcore/scanAccountsOnDevice.js
+++ b/src/libcore/scanAccountsOnDevice.js
@@ -16,6 +16,7 @@ import type {
 } from "@ledgerhq/live-common/lib/types";
 
 import { open } from "@ledgerhq/live-common/lib/hw";
+import Sentry from "react-native-sentry";
 import { withLibcoreF } from "./access";
 import { shouldShowNewAccount } from "../cryptocurrencies";
 import { syncCoreAccount } from "./syncAccount";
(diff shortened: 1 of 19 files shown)


src/libcore/scanAccountsOnDevice.js
```diff
@@ -93,7 +94,7 @@ export const scanAccountsOnDevice = (
         o.complete();
       } catch (e) {
         const mappedError = remapLibcoreErrors(e);
-
+        Sentry.captureException(mappedError);
```
gre: prefer passing the `e` that is the raw error

src/screens/SendFunds/06-Validation.js
```diff
@@ -109,6 +110,8 @@ class Validation extends Component<Props, State> {
           if (e && e.statusCode === 0x6985) {
             error = new UserRefusedOnDevice();
           }
+
+          Sentry.captureException(error);
```
gre: it should be in the `else` case of the if above. `UserRefusedOnDevice ` is legit

gre:
all good! some minor changes request.

src/libcore/scanAccountsOnDevice.js
juan-cortes: On it!

juan-cortes:
Done @gre

PR #785, opened 2019-02-22
LL-1072 Disable input of non ascii character for device name

juan-cortes:
To prevent characters that can't be displayed on the NanoX screen this adds a filter of non-ASCII characters to the edit device name screen. The behavior is a bit glitchy on Android where it will allow you to enter the character but clean-up only after another character is entered or we try to save. This is an issue with the events triggered on Android, works well on iOS.

Content is sanitized on both platforms. Try entering non-ASCII characters.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -33,6 +33,7 @@
     "lines-between-class-members": 0,
     "flowtype/space-after-type-colon": 0,
     "no-continue": 0,
+    "no-control-regex": 0,
     "no-return-assign": 0,
     "no-shadow": 0,
     "prefer-template": 0,
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     }
   },
   "dependencies": {
-    "@ledgerhq/errors": "4.39.0",
+    "@ledgerhq/errors": "4.41.1",
     "@ledgerhq/hw-app-btc": "4.38.0",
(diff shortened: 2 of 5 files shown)


juan-cortes:
I can't hide anything from the keyboard. I can use a validation instead of
removing the invalid on save, I mentioned it was glitchy in that way.

If you think disabling save and showing a warning makes more sense, I can
do it

Juan Cortés

gre:
There is no easy fix for this yeah. but to hope the production app will be fast enough.
I don't think there are any keyboard layout that would satisfy our need neither.

gre:
Ok DS is telling me the best idea is to go for a red validation under the field

gre:
So no on the fly validation but an error and wording that explains something

juan-cortes:
gotcha

Juan Cortés

juan-cortes:
Depends on ledgerhw/errors being updated to include the new error. I've added a modification to the remaining character count, it used to count the bytes taken and we could have negative values for the remaining characters. Now, since we don't allow non-ascii, all characters take one space.

image

juan-cortes:
Updated the ledgerhq/errors version so this pr should be ready for testing @Arnaud97234

juan-cortes:
Added the empty name (only spaces) to disable the button @Arnaud97234

PR #792, opened 2019-02-26
LL-1086 Handle new react-native-hid logic changes

juan-cortes:
Linked to https://github.com/LedgerHQ/ledgerjs/pull/303 , ~once merged react-native-hid will broadcast a `reset` event that we need to handle on the mobile end to clear the device list. This is that handling.~ What do we do about the unsubscribe logic that was unnecessary after the changes, do we remove it from here?

### DIFF
--- a/src/components/SelectDevice/index.js
+++ b/src/components/SelectDevice/index.js
@@ -92,7 +92,9 @@ class SelectDevice extends Component<Props, State> {
                   name: e.name,
                   family: e.family,
                 })
-              : devices.filter(d => d.id !== e.id),
+              : e.type === "reset"
+                ? []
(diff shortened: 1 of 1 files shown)


gre:
we'll re-discuss about this but i'm not sure about this semantic.

to me it should `"remove"` all the past emitted `"add"` if it's what a `"reset"` means.

juan-cortes:
After the discussion, there'll be no `reset` event since we can just delete the device, single device otg support for now since I can't test multiples.

juan-cortes:
I'm concerned here @gre if we remove the unsubscribe from mobile too, will it break the ble part?

PR #796, opened 2019-03-01
Fixes regression introduced by #784

gre:
some errors has been silently because buggy code introduced in #784 . to avoid this in future, we are factorizing this code in `logger.critical` code sample.

this was preventing the genuine check error to display at all! I think there are potentially other spots that was affected by this bug (basically instead of seeing the error, you still see things in pending state)

### Type

bug fix

### DIFF
--- a/scripts/post.sh
+++ b/scripts/post.sh
@@ -6,7 +6,7 @@ if [ -e node_modules/react-native-camera/ios/FaceDetector ] ; then
 fi
 cp node_modules/react-native-camera/postinstall_project/projectWithoutFaceDetection.pbxproj node_modules/react-native-camera/ios/RNCamera.xcodeproj/project.pbxproj
 
-rm -f 'node_modules/@segment/analytics-ios/.clang-format'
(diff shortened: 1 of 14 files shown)


juan-cortes:
Accepting this regression comes from my pr, how is this an issue at all? `Sentry` is the default export for the library, so `import {Sentry}` `import Sentry` and `import {default as Sentry}` should do the same thing? I'm confused @gre

gre:
Seems like Sentry is not exported as "default" from what I could tell

`TypeError: undefined is not an object (evaluating 'c.default.captureException')`

was the error and was for instance silenting the Genuine Check error – fixed after this

```
import { Sentry } from "react-native-sentry";
```

seems to be the correct syntax

if doing

```
import A from "react-native-sentry";
```

Sentry would be in `A.Sentry`

juan-cortes:
I should've tested it by forcing a crash, I don't think having extra reviews would've helped to catch this... Look at https://github.com/getsentry/react-native-sentry/blob/master/lib/Sentry.js default export is `Sentry` and doing `{Sentry}` in this case should (obviously not because it broke?) return the same thing

PR #798, opened 2019-03-04
[Android] Nano S support

juan-cortes:
image

`LL-*`

### Developer notes

Addition of two new env variables to chose what is the model id of the httpdebug (or the mock devices):

```
FALLBACK_DEVICE_MODEL_ID=nanoS
FALLBACK_DEVICE_WIRED=YES
```


gre:
- title to be in or out of the DeviceSelect ?
- scrollview should probably be IN the DeviceSelect because it's not only on Manager we'll need it ?
- polish: some unecessary `<View>` can be replaced by `<>`

gre:
team work 🚀

juan-cortes:
I'll take care of them

juan-cortes:
Latest version @Arnaud97234 solves both the disconnect/connect issue and the Unpair button that wasn't supposed to be there. We still have a blink (disconnect/connect) in the background that we'll need to address eventually. I've created an issue for it.

@gre there was no other way that I could find of getting this line to pass ci, sorry for the eyesore.

gre:
we'll try to see if it's possible to fix the "asking again" problem but AFAIK it's tricky. Not considering that blocking for merge, let's merge!

juan-cortes:
Did we not fix this if upon the first plugin of the device we accept the "launch this app when we plug the device" prompt? I don't get that prompt ever because I accepted it

gre:
@juan-cortes  yeah i think if you check that fix it!

PR #805, opened 2019-03-11
Fixes/20190311

gre:
details are in the commits


gre:
Everything should be fixed now.

src/components/AccountGraphCard.js
```diff
@@ -87,7 +87,7 @@ class AccountGraphCard extends PureComponent<Props, State> {
           isInteractive={isAvailable}
           isLoading={!isAvailable}
           height={100}
-          width={Dimensions.get("window").width - 40}
+          width={Dimensions.get("window").width - 32}
```
juan-cortes: Shouldn't this use `getWindowDimenions`? 
There's some platform specific stuff for android 🤷‍♂️

src/components/GraphCard.js
```diff
@@ -89,7 +89,7 @@ class GraphCard extends PureComponent<Props, State> {
           isInteractive={isAvailable}
           isLoading={!isAvailable}
           height={100}
-          width={getWindowDimensions().width - 40}
+          width={getWindowDimensions().width - 32}
```
juan-cortes: same

gre: Here it's used right?

gre:
@Arnaud97234 ready for review. the error cases are potential not fixed but it should not be a regression because i'm pretty sure it's as on develop.

PR #806, opened 2019-03-11
Reintegrated the readonly flag for iOS

juan-cortes:
Integrated back the readonlymode flag, for iOS.
Should show the pitch nanoX screen for iOS devices until we pair one, should be transparent for Android.

### DIFF
--- a/src/screens/Manager/ReadOnlyNanoX.js
+++ b/src/screens/Manager/ReadOnlyNanoX.js
@@ -1,6 +1,6 @@
 /* @flow */
 import React, { PureComponent } from "react";
-import { StyleSheet, View, Image, Linking } from "react-native";
+import { StyleSheet, View, Linking } from "react-native";
 import { Trans } from "react-i18next";
 import colors from "../../colors";
 import TrackScreen from "../../analytics/TrackScreen";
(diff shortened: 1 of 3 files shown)


src/screens/Manager/ReadOnlyNanoX.js
```diff
@@ -35,8 +39,9 @@ class ReadOnlyNanoX extends PureComponent<OnboardingStepProps> {
       <View style={styles.root}>
         <TrackScreen category="Manager" name="ReadOnlyNanoX" />
         <View style={styles.image}>
-          <Image source={require("../../images/nanoXHorizontalBig.png")} />
+          <DeviceNanoAction screen="empty" width={width * 0.8} />
```
gre: minor feedback but ideally we should explicitly set `deviceModel="nanoX"`

juan-cortes: After the last version from val https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/components/DeviceNanoAction.js#L21 it's the default prop though

gre: i was thinking that not providing it means "i don't care what device it is, give me a generic UI", but here we probably really want a NanoX

ok for now

PR #810, opened 2019-03-12
Fixes the Nano S issues on Onboarding

valpinkman:
LL-1152


gre:
maybe we need to set it to `2 / PixelRatio.get()` so it's exactly 2 pixels. because i think we're in point domain here.

gre:
We also need a new graphics.

src/components/SelectDevice/index.js
```diff
@@ -201,18 +214,23 @@ class SelectDevice extends Component<OwnProps, State> {
     );
 
     const hasUSBSection = Platform.OS === "android" || other.length > 0;
+    const isNanoSOnOnboarding = isNanoS && onboarding;
 
     return (
       <View>
-        {ble.length === 0 ? (
+        {isNanoS && onboarding ? (
```
gre: I think the condition can just be `isNanoS` , you can also remove `isNanoSOnOnboarding`

src/screens/Onboarding/steps/pair-new.js
```diff
@@ -28,16 +28,23 @@ class OnboardingStepPairNew extends Component<OnboardingStepProps> {
   pairNew = () => this.props.navigation.navigate("PairDevices");
 
   render() {
+    const { deviceModelId, t } = this.props;
+    const isNanoS = deviceModelId === "nanoS";
```
gre: **/!\** it does not include the Blue and should!

( rename it to usbOnly )

gre:
after Valentin fixes this, reminder that we need to test with a Blue, (the design won't match, the wording should tho)

valpinkman:
@gre Left To Do

- Verify address message still shows wrong text
- Started to contextualize the DeviceJob with deviceModelId, but did not integrate the textes or the correct illustration

I think I've covered the other cases

gre:
Edit Device name exists only for NanoX over BLE

gre:
@Arnaud97234 ready for review

PR #817, opened 2019-03-14
LL-1066 Fixes scaling issues in the app

juan-cortes:
It was a bit more involved than initially thought but straightforward enough, this pr fixes (or attempts to do so) the scaling issues when changing the font size scale on the system level. Ideally this should be a temporary thing and we should support some sort of font change to allow the user to adapt the font size.

### How to test
Change the scaling value for both platforms and go over all screens and flows.

### DIFF
--- a/src/components/CurrencyInput.js
+++ b/src/components/CurrencyInput.js
@@ -142,6 +142,7 @@ class CurrencyInput extends PureComponent<Props, State> {
     return (
       <View style={[styles.wrapper, style]}>
         <TextInput
+          allowFontScaling={false}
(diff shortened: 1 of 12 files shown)


gre:
beautiful

juan-cortes:
Ok, so iphone se issues should be fixed, and that applies to also android (the scanning progress). I had to hack the default props of `Text` for now, until a pr is accepted from the progress circle library, I'll send one now.

As for the samsung galaxy s9.... can't replicate, no idea, can't fix it right now.

PR #820, opened 2019-03-17
Added the countervalue field for fees on operation details

juan-cortes:
image

🤷‍♂️

### DIFF
--- a/src/components/CounterValue.js
+++ b/src/components/CounterValue.js
@@ -31,6 +31,7 @@ type OwnProps = {
   // as we can't render View inside Text, provide ability to pass
   // wrapper component from outside
   Wrapper?: React$ComponentType<*>,
+  subMagnitude?: number,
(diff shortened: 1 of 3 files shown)


gre:
To be checked with design but I imagine it could be possible to put it next to it maybe in a greyish style in parenthesis. I also think we could try to turn on the "subMagnitude" to 3 (or something like that) so the thing is able to display 0.00001 USD in this extreme case.

juan-cortes:
This gives design something to work from, I don't mind _wasting_ the 2 minutes it took

juan-cortes:
- Increased the precision of the counter values
- Made the counter value use the rate of the operation date instead of the latest
- Updated the design with @dekkaki feedback
- Changed the default color for android on the values, it was getting `smoke` and wanted `darkblue` or similar, to match iOS.

image

PR #822, opened 2019-03-18
LL-1184 Pixel push mobile 2

juan-cortes:
image

- Removed the small add button and replaced it with the old full row one
- Replaced "Connect" with "Select", on Manager
- I can't get the ellipsis to be blurry @gre @aramab , lgtm
- Increased the size of the illustration to pair a new nano X

### DIFF
--- a/src/components/SelectDevice/PairNewDeviceButton.js
+++ b/src/components/SelectDevice/PairNewDeviceButton.js
@@ -1,12 +1,13 @@
 // @flow
 
 import React, { PureComponent } from "react";
-import { View, StyleSheet, TouchableOpacity } from "react-native";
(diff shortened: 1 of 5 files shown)


gre:
In term of wording, I would not use the word "Pair". Maybe just "new Ledger Nano X" ?

juan-cortes:
> In term of wording, I would not use the word "Pair". Maybe just "new Ledger Nano X" ?

If you are sure, I'll change it. Should we invoke the word genie?  @dasilvarosa

gre:
can you 

```
    "deviceNotFoundPairNewDevice": "Add new Ledger Nano X",
```

?

juan-cortes:
Done @gre

PR #830, opened 2019-03-21
improve mock

gre:
to be continued:

- [x] pair a new device
- [x] enter manager
- [x] the Send flow should work up to the end.
  - [x] we need to bypass the device step in some way. it can just be the present of a NEXT button in the mock mode.
  - [x] we need to fix visiblity of all fees / specific fields. (=> XRP, ETH, BTC all need to mock the fee / custom fields so they are fully functional)
  - [x] we need to cover the validation of all fields. (=> for instance we need a way to test mock validation of address field (it can just be as dumb as there must be at least 3 characters, just so we can test the 2 UI cases))

### Future

we will polish these in the future:

- making sure all the async features can be interrupted with an error. Use `rejectionOp(observable)` / `hookRejections(promise)` to do that easily. Make sure a DebugRejectSwitch button is always visible on screen. If not, we can fix its visibility.
- it is possible we might want to implement more advanced stuff than just "one big stop button". E.G. for validation on device stuff, we might want to display in our mock UI a way to "validate/deny" on device.. but overkill for now!

### Type

QA & prepare for test automation

**The most important thing to test is to make sure the app works normally without the MOCK mode**, testing the mock itself is not really important and will be part of waldo.io anyway.

If still want to test the mock:

need to use the env of `.env.mock`

Then you should be able to browse to all part of the app without blocker. Add a new device (mock of ble), add accounts, receive, send, manager,... all these part are mocked. Some feature are obviously not working like Install/Uninstall apps but it's acceptable as first MVP.


package.json
```diff
@@ -41,7 +43,7 @@
     "@ledgerhq/hw-app-xrp": "4.48.0",
     "@ledgerhq/hw-transport": "4.48.0",
     "@ledgerhq/hw-transport-http": "4.48.0",
-    "@ledgerhq/live-common": "4.34.1",
+    "@ledgerhq/live-common": "4.35.0",
```
valpinkman: I've already merged a branch with `live-common` v4.36.1, be careful not to overwrite it

PR #832, opened 2019-03-25
Clear input button on recipient address, send flow, max cap

juan-cortes:
- fix unlinked clear button on input for the recipient address
- adds a max fee cap to custom fees for bitcoin-like coins


Luuk2016:
Great!

PR #849, opened 2019-04-02
Bump live-common with getDeviceInfo fix

gre:


src/index.js
```diff
@@ -29,6 +29,8 @@ import HookAnalytics from "./analytics/HookAnalytics";
 import HookSentry from "./components/HookSentry";
 import AppContainer from "./navigators";
 
+console.log(NativeModules)
```
juan-cortes: what is this?

gre: oops

PR #850, opened 2019-04-02
feat: LL-1077 Add failsafe to the send flow when fees are too high

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1077

The send flow, try to send an amount with fees representing more than 10% of the amount, you should be prompted with the above modal to confirm that unusual transaction. ~~The modal should not be dismissable by clicking outside~~ the user needs to either confirm and go to the next step, or cancel and remain on the first step of the flow.

### DIFF
--- a/src/icons/AlertTriangle.js
+++ b/src/icons/AlertTriangle.js
@@ -1,13 +1,17 @@
 // @flow
 import React from "react";
 import Svg, { Path, G } from "react-native-svg";
+import colors from "../colors";
 
 type Props = {
   size: number,
   color: string,
 };
 
-export default function AlertTriangle({ size = 16, color }: Props) {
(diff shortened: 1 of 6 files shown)


src/icons/AlertTriangle.js
```diff
@@ -7,7 +7,7 @@ type Props = {
   color: string,
 };
 
-export default function AlertTriangle({ size = 16, color }: Props) {
+export default function AlertTriangle({ size = 16, color = "#ea2e49" }: Props) {
```
gre: not a `colors` constant?

juan-cortes: done sir

PR #861, opened 2019-04-10
add experimental feature in settings

valpinkman:
Adds a new **Experimental Features** section in settings similar to LLD

### Type

Feature

### Context

LL-1288

Settings

image

image


ios/ledgerlivemobile.xcodeproj/project.pbxproj
```diff
@@ -598,13 +598,6 @@
 			remoteGlobalIDString = 274692C321B4414400BF91A8;
 			remoteInfo = "RNSentry-tvOS";
 		};
-		F8A31E45223A3BDF00F81049 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = 64C6A31B337D471B93E32DCD /* RNSentry.xcodeproj */;
-			proxyType = 2;
-			remoteGlobalIDString = 274692C321B4414400BF91A8;
-			remoteInfo = "RNSentry-tvOS";
-		};
```
gre: was this accidental?

src/experimental.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  for (const k in Config) {
+    setEnvUnsafe(k, Config[k]);
+  }
+  /* eslint-enable guard-for-in */
+
+  changes.subscribe(async ({ name, value }) => {
+    if (experimentalFeatures.find(f => f.name === name) && !isReadOnly(name)) {
+      await setStorageEnvs(name, value);
```
gre: there are potential race condition here, an event can comes in the stream before the previous one to be awaited, making the first set failing because you getStorageEnv of the old envs on the second call.

a solution could be to use a `concatMap` and returning a `from(promise)` , but basically the current code being async/await does not solve the issue and it's the same as not awaiting.

I would suggest you just have a `setStorageEnvs()` without params that will just save from what is in the env (but i guess only the subset that is here?)

src/logic/withEnv.js
```diff
@@ -0,0 +1,47 @@
...
+
+    componentWillUnmount() {
+      if (this.sub) {
+        this.sub.unsubscribe();
+      }
+    }
+
+    init = () => {
+      const envs = getAllEnvs();
+      this.setState({ env: envs[name] });
```
gre: we shouldn't do a setState at init, it should directly be in the state initially, this would avoid a double rendering.

src/logic/withEnvs.js
```diff
@@ -0,0 +1,42 @@
...
+
+    componentWillUnmount() {
+      if (this.sub) {
+        this.sub.unsubscribe();
+      }
+    }
+
+    init = () => {
+      const envs = getAllEnvs();
+      this.setState({ envs });
```
gre: same `this.state = { envs: getAllEnvs() }`

src/experimental.js
```diff
@@ -0,0 +1,106 @@
...
+    title: "Dev mode",
+    description: "enables developer apps in manager",
+  },
+  // {
+  //   type: "toggle",
+  //   name: "EXPERIMENTAL_USB",
+  //   title: "Experimental USB",
+  //   description:
+  //     "Alternative USB implementation that might help solving USB issues. Enabling this feature might create UI glitches.",
+  // },
```
gre: remove the EXPERIMENTAL_USB thing, it will never exist on mobile

src/experimental.js
```diff
@@ -0,0 +1,106 @@
...
+  // {
+  //   shadow: true,
+  //   type: "toggle",
+  //   name: "FORCE_PROVIDER",
+  //   valueOn: 4,
+  //   valueOff: 1,
+  //   title: "Manager test app provider=4",
+  //   description: "enables yet `unreleased` apps in manager",
+  // },
```
gre: since you still have the shadow system, i guess you can uncomment it ;) i think it still makes sense when we do custom build to have this feature ;)

ios/ledgerlivemobile.xcodeproj/project.pbxproj
valpinkman: I used add -p but wasn't sure since I reinstalled packages. for now I guess I'll reset it to `develop`'s state

PR #862, opened 2019-04-10
feat: User should be able to edit account name when adding an account

juan-cortes:


src/components/SelectableAccountsList.js
```diff
@@ -112,17 +225,34 @@ class SelectableAccount extends PureComponent<{
         <AccountCard account={account} />
         {!isDisabled && (
           <CheckBox
+            onChange={onPress?() =>onPress(account):undefined}
```
gre: `onPress ? this.onPress : ..`

src/components/SelectableAccountsList.js
```diff
@@ -88,20 +114,107 @@ class SelectableAccountsList extends Component<ListProps> {
...
+    const { rowIndex, listIndex } = this.props;
+    this.panResponder = PanResponder.create({
+      // Ask to be the responder:
+      onStartShouldSetPanResponder: () => true,
+      onStartShouldSetPanResponderCapture: () => false,
+      onMoveShouldSetPanResponder: () => false,
+      onMoveShouldSetPanResponderCapture: () => false,
+
+      onPanResponderGrant: () => {
+        if (swipedAccountSubject) {
```
gre: Why the if?

src/components/SelectableAccountsList.js
juan-cortes: how would that pass the account though?

src/components/SelectableAccountsList.js
juan-cortes: I was getting undefined if it triggered while we are still scanning, probably something like the component is there but the ref is not ready yet... I don't know. I needed it.

gre: Isn't that variable a global immutable subject? So it's always defined

juan-cortes: don't know what to tell you, if I remove the check it sometimes crashes

PR #864, opened 2019-04-12
feat: Add support for adaptive launcher icons on Android v26+

juan-cortes:
In order to follow the [guidelines for adaptive icons](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive) we need to provide the launcher icon as a two-layered drawable. This should only affect Android 26+ and it can be tested by changing the icon shape in your system. 

NB_Icon_Layers_3D_03_ext

### Type

Feature

### Context

Without this, since we provide a single layer drawable. Launchers with a white background would show the LLM logo without a shape surrounding it. This worked well for lower versions and for launchers without a white background.  👎 
image

### After this is merged
We will allow users to change the shape of their icons like they can do for other supporting apps. Note that the icons in the following screenshot dont look completely aligned, it's a glitch in the sketch file used to generate them.

image


gre:
:mindblown:

PR #866, opened 2019-04-15
fix: Addresses Samsung inconsistencies and fontScaling issues

juan-cortes:
For larger font sizes we had two remaining issues that were mainly present on Samsung devices. First of all, the footer modals that had bullet point lists were shown with huge texts that broke the design. Since this is the first approach at accessibility where we stop the font change altogether to have a stable base for future accessibility we needed to address that.

The issue was caused by[ nested text components](https://github.com/facebook/react-native/issues/19833), causing the `allowFontScaling` prop to be lost at the parent level and rendering the child elements, in this case, the text points of the list, with the scaled size.

The second error was affecting only Samsung devices with font scaling and it made password inputs unusable. This is caused by not using our wrapper for text input and relying on the native one that doesn't have the font scaling recalculation. That is also solved in this pr.

### Type

Fix

### Context

LL-1178

Same as the other time we had a font scaling test plan, go over the app with scale mode enabled and see if there are more broken things. There is one known issue that I've been unable to fix so far, but it's so much more edge case than before that I think we can safely ignore it for now.

### Known issue
With font scaling enabled, on Android, on a text input that allows you to clear the input, the font size goes pretty small (half size) after clearing with the Right X button.

### DIFF
--- a/src/components/TextInput.ios.js
+++ b/src/components/TextInput.ios.js
@@ -6,7 +6,8 @@ import React, { PureComponent } from "react";
 class TextInput extends PureComponent<*> {
   render() {
     const {
-      containerStyle, // Needed to pass flow, since we call the native TextInput
+      containerStyle, // For Android it's two different styles, here we join them
+      style,
       withSuggestions,
       innerRef,
       ...otherProps
@@ -22,6 +23,7 @@ class TextInput extends PureComponent<*> {
       <ReactNativeTextInput
         ref={innerRef}
         allowFontScaling={false}
+        style={[containerStyle, style]}
         {...otherProps}
         {...flags}
       />
--- a/src/index.js
+++ b/src/index.js
@@ -38,11 +38,11 @@ const styles = StyleSheet.create({
 
 // Fixme until third parties address this themselves
 // $FlowFixMe
-Text.defaultProps = { ...(Text.defaultProps || {}), allowFontScaling: false };
+Text.defaultProps = { ...Text.defaultProps, allowFontScaling: false };
 // $FlowFixMe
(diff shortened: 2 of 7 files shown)


src/components/TextInput.ios.js
```diff
@@ -22,6 +23,7 @@ class TextInput extends PureComponent<*> {
       <ReactNativeTextInput
         ref={innerRef}
         allowFontScaling={false}
+        style={[containerStyle, style]}
```
gre: this PR fixes Samsung thing, why is there a change needed on iOS impl too?

src/index.js
```diff
@@ -38,9 +38,14 @@ const styles = StyleSheet.create({
 
 // Fixme until third parties address this themselves
 // $FlowFixMe
-Text.defaultProps = Text.defaultProps || {};
+Text.defaultProps = { ...(Text.defaultProps || {}), allowFontScaling: false };
```
gre:
```suggestion
Text.defaultProps = { ...Text.defaultProps, allowFontScaling: false };
```

src/index.js
```diff
@@ -38,9 +38,14 @@ const styles = StyleSheet.create({
...
 // $FlowFixMe
-Text.defaultProps.allowFontScaling = false;
+TextInput.defaultProps = {
+  // $FlowFixMe
+  ...(TextInput.defaultProps || {}),
```
gre:
```suggestion
  ...TextInput.defaultProps,
```

gre:
need more test on all phones

juan-cortes:
we might revisit this, it's an eye sore seeing it unmerged :trollface:

PR #868, opened 2019-04-16
Bump live-common

gre:
like https://github.com/LedgerHQ/ledger-live-desktop/pull/1936 but on mobile


ios/ledgerlivemobile.xcodeproj/project.pbxproj
```diff
@@ -1371,6 +1364,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
```
valpinkman: same changes that I had O_O

gre: I guess it's ok

gre:
@Arnaud97234 these are invalid path and was intentionally removed from scanning on mobile as we chat about it. If users need to recover invalid path funds, they will need the Desktop app.

i'm fixing the libcore issue on the other hand.

Arnaud97234:
> @Arnaud97234 these are invalid path and was intentionally removed from scanning on mobile as we chat about it. If users need to recover invalid path funds, they will need the Desktop app.
> 
> i'm fixing the libcore issue on the other hand.

PR #871, opened 2019-04-18
replace seVersion by fullVersion

gre:
In incoming refactoring, deviceInfo.seVersion will be gone


gre:
i'm going to address this in the live-common PR

PR #875, opened 2019-04-18
Fixes crash when deleting an account

gre:
it should fix the crash when delete an account but i've not effectively test it.


gre:
@juan-cortes could you confirm this PR fixes it? I'm busy on the 2 other PRs that requires 100% of my CPU to compute the apps xD

juan-cortes:
Late and already merged, but yes, this seems to fix it.

PR #884, opened 2019-04-24
LL-1320 Adds missing styles for unconfirmed operations

juan-cortes:
Adding the unconfirmed state to operations, both on the operation list and operation details while at the same time reworking the SVG icons (thanks @bscholving) since there were some issues with masking and react native.
image

### Type

Feature, UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1320

Operation list row and Operation details for confirmed/unconfirmed operations


src/components/OperationRow.js
```diff
@@ -1,36 +1,45 @@
...
 };
 
 const placeholderProps = {
   width: 40,
   containerHeight: 20,
 };
 
+const mapStateToProps = createStructuredSelector({
+  currencySettings: currencySettingsForAccountSelector,
+});
```
gre: I'm a bit worried about overall performance of this. like we did on desktop, I would really prefer we don't connect the full row but that we make a dedicated `OperationConfirmationIcon` component that is connected (that would be an intermediary light component). that way, we avoid unecessary redraw each time the redux updates (which is after each sync for instance)

juan-cortes: 👍 The current version should address that, minimizing the redraw

juan-cortes:
@Arnaud97234

PR #885, opened 2019-04-24
UI Polish Simplified Confetti, alternative approach

juan-cortes:
There is a pretty obvious lag between the `share analytics` step and the `finish` step of the onboarding flow. This made the app seem irresponsive for a few seconds, depending on the speed of your device. I played with a few alternatives like using pngs, prerendering on previous steps, not using svgs, etc but the performance gains were negligible, at least from my vague measurements.

This is a compromise where we will render the page immediately after the user clicks on the button by using `InteractionManager.runAfterInteractions` and then in the background try to do the confetti party without making the app choke. Granted the confetti takes a while to appear, but I'd rather be able to ignore the confetti and Open the app, and be surprised by the confetti if I'm in that screen long enough than be stuck in a white screen until the forced confetti appears.

I removed all the `emit` logic that we aren't currently using too, it's still in desktop shall we want to randomly emit particles somewhere :trollface: 

### Type

Performance, UI Polish

Confetti party after onboarding.

### DIFF
--- a/src/components/ConfettiParty/index.js
+++ b/src/components/ConfettiParty/index.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React, { PureComponent } from "react";
-import { View } from "react-native";
+import { View, InteractionManager } from "react-native";
 import Confetti1 from "../../icons/confetti/confetti1";
 import Confetti2 from "../../icons/confetti/confetti2";
 import Confetti3 from "../../icons/confetti/confetti3";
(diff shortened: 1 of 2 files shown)


src/components/ConfettiParty/index.js
```diff
@@ -1,107 +1,60 @@
...
-      }, 40);
-    } else {
-      clearInterval(this.interval);
-    }
-  }
-  interval: *;
-  initialInterval: *;
-  initialTimeout: *;
+  emitConfetti = () => {
+    setTimeout(
```
gre: Do you really need this setTimeout? runAfterInteraction is definitely a good idea and running in that I don't think you longer need this timeout.

src/components/ConfettiParty/index.js
```diff
@@ -1,107 +1,60 @@
...
+  };
 
   render() {
-    const { confettis } = this.state;
     return (
       <View style={{ position: "relative", width: "100%", height: "100%" }}>
-        {confettis.map(c => (
-          <Confetti key={c.id} {...c} />
-        ))}
+        {this.state.confetti.map(C => C)}
```
gre: map not required?

src/components/ConfettiParty/index.js
juan-cortes: My initial findings showed a marginal improvement, probably shadowed after being inside the runafterinteractions. Removed in current version

src/components/ConfettiParty/index.js
juan-cortes: indeed. Removed in current version

src/components/ConfettiParty/index.js
```diff
@@ -1,107 +1,61 @@
...
+    this.handler = InteractionManager.runAfterInteractions(() => {
+      this.emitConfetti();
+    });
   }
 
   componentWillUnmount() {
-    this.setEmit(false);
-    clearInterval(this.initialInterval);
-    clearTimeout(this.initialTimeout);
+    this.handler.cancel();
```
valpinkman: maybe check first

```
if (this.handler) this.handler.cancel()
```

juan-cortes: (ﾉ≧∇≦)ﾉ ﾐ ┸━┸

valpinkman: (╯°□°)╯︵ ┻━┻ ┻━┻ ︵ ヽ(°□°ヽ)

juan-cortes: Done (ヘ･_･)ヘ┳━┳

PR #886, opened 2019-04-26
Migrate to React Native 59

gre:
following http://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059

- [x] ~~can we drop dep to jsc?~~ => NO, jsc is still older on RN.
- [x] ~~metro config js => should we try to enable something?~~ => not for now
- [x] replacing deprecated RN feature to use libs
- [x] for Android, some more stuff to patch from this diff: https://github.com/react-native-community/rn-diff-purge/compare/release/0.58.6..release/0.59.5#diff-b5a6cecd7ee362715d6306d583e73612R116
- [x] fixes an icon glitch on Android

To be tested:

- [x] it works in simulator iOS
- [x] it works in simulator Android
- [x] it works in iOS phone
- [x] it works in Android phone
- [ ] it Archive on iOS
- [ ] it build and deploy in Release on iOS phone.
- [x] a staging-android works and deploy correctly to a phone

We also need to not forget about:

- check analytics still works
  - [ ] iOS
  - [ ] Android
- check Sentry still works
  - [ ] iOS
  - [ ] Android


android/app/build.gradle
```diff
@@ -180,6 +180,7 @@ configurations.all {
 }
 
 dependencies {
+    implementation project(':@react-native-community_netinfo')
     compile project(':lottie-react-native')
```
gre: @MortalKastor  I wonder if we need to replace `compile` by `implementation` like it seems to have be done when i've used `react-native link`

juan-cortes: I played with that last week, I thought I actually made a pr about it... replacing all compiles with implementation made my build times faster and gradle happier.

juan-cortes: I was sure I replied to this... I made a test and apparently never made the pr, but updating all compile to implementation sped up my app launches quite a bit, and made gradle happier

MortalKastor: I don't see any change myself, but at least it doesn't seem to break anything ¯\_(ツ)_/¯

gre: 👏

gre:
the bug shared above is present on develop and we'll fix it there

PR #888, opened 2019-04-29
Use mocked countervalues from common if MOCK

juan-cortes:
Use countervalue mocks if we are using the MOCK flag. 

### Type

Testing Feature

### Context

This is not complete because the usage of `timemachine` library resulted in an irresponsive app that started doing very strange things. This means that historical rates are faked, but the latest rate is still considering the latest time. I'm open to ideas on how to address this @gre 

~Also, flow is not happy with this version~ Flow is now happy


gre:
feature request to work with other than USD

PR #891, opened 2019-05-07
implement SkipLock to fixes LL-1341

gre:
it seems the component was there but never implemented?

- [ ] to test on iOS (real phone)
- [ ] to test on Android (real phone)

it should skip the phone going in sleep mode during critical actions like install apps, scan accounts, pairing screen, verify on device,...


LFBarreto:
i think this was fixed, closing this

PR #892, opened 2019-05-07
Fix connection to bundler on Android (in dev mode)

MortalKastor:
Starting with Android 9.0 (API level 28), cleartext communication is disabled by default, which breaks the plain HTTP connection to the bundler used in dev mode.

This PR allows un-encrypted traffic for the app in dev mode *only*.

### TO DO

- [ ] Check if we rely on plain HTTP in production
- [ ] Whitelist domains in `android/app/src/main/AndroidManifest.xml` if we do


gre:
AHA 👌

MortalKastor:
I fear we'll be unable to use the debug HTTP transport in production builds as it is 😢

PR #902, opened 2019-05-14
adds failed status to eth sent transactions

valpinkman:
adds a failed status for ETH transactions in operations list and operation details

### Type

Improvement

### Context

LL-1316

Account -> Operations List and Operation Details

No screen as the EXPERIMENTAL_EXPLORERS are not working on mobile yet (getting a `Oops internet seems down` error) so it's an optimistic feature ¯\_(ツ)_/¯


src/components/OperationIcon.js
```diff
@@ -24,15 +25,24 @@ class OperationIcon extends PureComponent<Props> {
...
+    const {
+      type,
+      size,
+      confirmed,
+      operation: { hasFailed },
+    } = this.props;
+    const Icon = this.icons[type][hasFailed ? 2 : confirmed ? 1 : 0];
+
+    return (
+      <>
```
gre: This is not needed.

src/components/OperationIcon.js
```diff
@@ -24,15 +25,24 @@ class OperationIcon extends PureComponent<Props> {
...
-    const Icon = this.icons[type][confirmed ? 1 : 0];
-
-    return <Icon size={size} />;
+    const {
+      type,
+      size,
+      confirmed,
+      operation: { hasFailed },
+    } = this.props;
+    const Icon = this.icons[type][hasFailed ? 2 : confirmed ? 1 : 0];
```
gre: just in case `Icon` is falsy, follow this line by a `if (!Icon) return null;`

PR #908, opened 2019-05-21
LL-1387 Changed the look of the UserRefusedAllowManager error

juan-cortes:
I didn't add the close button because I didn't see an easy way of adding it, if it's a big deal (you can dismiss and close with the X top right) I can dedicate more time to it.

image

### Type

UI Polish
### Context

https://ledgerhq.atlassian.net/browse/LL-1387

Inside the manager, when selecting a device, refuse the authorization on the device. It should trigger the above error instead of the one shown in the issue. We could potentially show a different device icon depending on the device since we can refuse on both types of devices... but I think it's good for now.

### DIFF
--- a/src/components/DeviceJob/StepRunnerModal.js
+++ b/src/components/DeviceJob/StepRunnerModal.js
@@ -28,7 +28,7 @@ class SelectDeviceConnectModal extends PureComponent<{
         id="DeviceJobModal"
         isOpened={!!meta}
         onClose={onClose}
-        preventBackdropClick
+        preventBackdropClick={error ? undefined : true}
       >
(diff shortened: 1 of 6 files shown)


gre:
@juan-cortes i also noticed in such case we also can't dismiss by click on black overlay anymore. I feel it's a bit extreme, only the pending should be prevented.

juan-cortes:
> @juan-cortes i also noticed in such case we also can't dismiss by click on black overlay anymore. I feel it's a bit extreme, only the pending should be prevented.

This is probably because of the thing we did to device steps right? I'll take a look, this error should be dismissable

Issue #263, opened 2018-10-01
ios: libcore support: implement a switch to support both Sim' and Device at same time


gre:
- the universal build must not contain x86 because it create problem at apple submission
- the simulator need a x86 build to be able to run. we need it for dev.

solution:

- use Header search and specify per distribution (sim vs ios device) the search headers for the dylib
- a prebuild script that will copy the correct dylib to the build. (either the x86 one or the universal one)

PR #915, opened 2019-05-27
[iOS] Strip x86 libcore binaries on "archive"

MortalKastor:
x86_64 libcore binaries needed for the iPhone simulator where embedded when doing an _archive_ to deploy on App Store, which was getting promptly rejected unless removed manually.

This should fix that ⓒ 

#### Before

Screen Shot 2019-05-27 at 17 00 44

#### After

Screen Shot 2019-05-27 at 17 02 11

### Type

Bugfix

### Context

- https://ledgerhq.atlassian.net/browse/LL-916
- #263 

- Archive
- Publish
- ???
- Profit


gre:
w'll take this. building the app will see if there are any glitch but looks good to me.

PR #917, opened 2019-05-28
LL-1403 Asset Distribution screen

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1403

### Known issues/Blocks
~We need to agree on the behavior for when counter values are missing or broken, do we redirect back, do we show a placeholder.~ In any of these cases, the button to access the distribution screen will be gone.


src/components/CurrencyIcon.js
```diff
@@ -3,11 +3,10 @@
 import React, { PureComponent } from "react";
 import { View, StyleSheet } from "react-native";
 import { getCryptoCurrencyIcon } from "@ledgerhq/live-common/lib/reactNative";
-import type { CryptoCurrency } from "@ledgerhq/live-common/lib/types";
 import LText from "./LText";
 
 type Props = {
-  currency: CryptoCurrency,
+  currency: *,
```
gre: either we need to revert, or you can try to `currency: Currency` and handle more cases.
(in the future we'll need to implement it so it defaults to a rendering with the ticker anyway)

src/components/CurrencyRate.js
```diff
@@ -0,0 +1,51 @@
...
+    iconSize: 10,
+  };
+
+  render() {
+    const { currency, fontStyle, iconSize } = this.props;
+    const one = new BigNumber(10 ** currency.units[0].magnitude);
+    return (
+      <View style={styles.wrapper}>
+        <Icon name="activity" color={currency.color} size={iconSize} />
+        <LText tertiary style={[styles.text, fontStyle]}>
```
gre: set a `numberOfLines={1}`

src/components/CurrencyRate.js
```diff
@@ -0,0 +1,51 @@
...
+          {`${currency.ticker} 1 = `}
```
gre: uses a `<CurrencyUnitValue>` because the formatting changes depending on the languages.

src/navigators.js
```diff
@@ -110,6 +110,7 @@ import DisplayResult from "./screens/ImportAccounts/DisplayResult";
 import FallBackCameraScreen from "./screens/ImportAccounts/FallBackCameraScreen";
 import OnboardingOrNavigator from "./screens/OnboardingOrNavigator";
 import AdvancedLogs from "./screens/AccountSettings/AdvancedLogs";
+import DistributionScreen from "./screens/Distribution";
```
gre: (minor. can be named Distribution in the first place)

src/screens/Distribution/DistributionCard.js
```diff
@@ -0,0 +1,118 @@
...
+import colors from "../../colors";
+import LText from "../../components/LText";
+import CurrencyIcon from "../../components/CurrencyIcon";
+import CurrencyUnitValue from "../../components/CurrencyUnitValue";
+import ProgressBar from "../../components/ProgressBar";
+import CounterValue from "../../components/CounterValue";
+import CurrencyRate from "../../components/CurrencyRate";
+
+export type DistributionItem = {
+  currency: (CryptoCurrency | TokenCurrency) & { color: string },
```
gre: I prefer we stick to 

`currency: CryptoCurrency | TokenCurrency`

because this is getting ugly

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,70 @@
...
+import Svg, { Path, G } from "react-native-svg";
+import colors from "../../colors";
+import type { DistributionItem } from "./DistributionCard";
+
+type Props = {
+  data: *,
+  style?: *,
+};
+
+// TODO refactor this so it's reusable not only with asset distributions
```
gre: (i think it's fine to make it specific)

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,70 @@
...
+class RingChart extends PureComponent<Props> {
+  static defaultProps = {
+    data: [],
+  };
+  arcGenerator = d3shape.arc();
+  reducer = (data: *, item: DistributionItem) => {
+    const increment = item.distribution * 2 * Math.PI;
+    const done = data.done;
+
+    if (done) return data;
```
gre: I don't see where done is used. sounds like it can be removed?

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,70 @@
...
+  render() {
+    const { data, style } = this.props;
+    const paths = data.reduce(this.reducer, { items: [], angle: 0 });
+
+    return (
+      <View style={style}>
+        <Svg width="100%" height="100%" viewBox="0 0 100 100">
+          <G transform="translate(50, 50)">
+            {paths.items.map(({ pathData, color }) => (
+              <Path
```
gre: use a `key` for React

src/screens/Distribution/index.js
```diff
@@ -0,0 +1,203 @@
...
+  accounts: accountsSelector,
+});
+
+class Distribution extends PureComponent<Props, *> {
+  static navigationOptions = {
+    title: i18next.t("distribution.header"),
+    headerLeft: null,
+  };
+
+  renderItem = ({ item, _ }: { item: DistributionItem, index: number }) => (
```
gre:
```suggestion
  renderItem = ({ item }: { item: DistributionItem, index: number }) => (
```

 🤔

src/screens/Distribution/index.js
```diff
@@ -0,0 +1,203 @@
...
+                  unit={counterValueCurrency.units[0]}
+                  value={distribution.sum}
+                />
+              </LText>
+            </View>
+          </View>
+          <LText bold secondary style={styles.distributionTitle}>
+            <Trans i18nKey="distribution.list" />
+          </LText>
+          <FlatList
```
gre: the part above the FlatList could be put into a `HeaderComponent` so you don't need the top level ScrollView. that simplifies a bit the nesting and can be better for performance. like we did on Portfolio basically.

src/screens/Distribution/index.js
```diff
@@ -0,0 +1,203 @@
...
+    <DistributionCard item={item} />
+  );
+
+  keyExtractor = item => item.id;
+
+  render() {
+    const distribution: AssetsDistribution = getAssetsDistribution(
+      this.props.accounts,
+      this.props.getCounterValue,
+    );
```
gre: I suggest to move that directly in the selector code. so basically you don't send the `getCounterValue` function, instead you send the `distribution` object.

later we might implement cache in live-common but we can only benefit it if it's sent as a prop.

src/screens/Portfolio/AnimatedTopBar.js
```diff
@@ -16,14 +16,30 @@ import BalanceHeader from "./BalanceHeader";
...
+    navigation: *,
+  },
+  { skipRender: boolean },
+> {
+  state = {
+    skipRender: true,
+  };
+  componentDidMount() {
+    this.props.scrollY.addListener(({ value }) =>
+      this.setState({ skipRender: value < 90 }),
```
gre: This can be pretty bad for performance. this triggers a setState regardless if the value changes, which means potentially a lot of redraw.

```
setState({ skipRender: false })
setState({ skipRender: false })
setState({ skipRender: false })
setState({ skipRender: true })
setState({ skipRender: true })
setState({ skipRender: true })
```

would trigger 6 re-render()


What you can do instead is this:


```
this.setState(old => {
  const skipRender = value < 90;
  if (skipRender===old.skipRender) return null;
  return { skipRender };
});
```

src/screens/Portfolio/AnimatedTopBar.js
```diff
@@ -16,14 +16,30 @@ import BalanceHeader from "./BalanceHeader";
...
+class AnimatedTopBar extends PureComponent<
+  {
+    scrollY: AnimatedValue,
+    portfolio: Portfolio,
+    counterValueCurrency: Currency,
+    pending: boolean,
+    error: ?Error,
+    navigation: *,
+  },
+  { skipRender: boolean },
```
gre: can you explain / comment in this code, why we need this variable?

skipRender is not enough explicit to understand what it does (maybe should rename it too. it feels like you are actually skipping the render()ing which is not the case)

src/components/CurrencyIcon.js
juan-cortes: I had it with Currency but there's no color there and it also breaks the icon, this was the cleanest even if not the most flow happy

src/components/CurrencyRate.js
juan-cortes: ok

src/components/CurrencyIcon.js
gre: I think we can use `Currency` and put a FlowFixMe.

i'll discuss with flow folks to see if there is a solution for our problem, idk why Flow struggle solving this.

src/components/CurrencyRate.js
juan-cortes: oh i see

src/navigators.js
juan-cortes: Comes from the fact that I had that superflous stack, changed

src/screens/Distribution/index.js
juan-cortes: 😄

src/screens/Distribution/index.js
juan-cortes: Oh yeah, you mentioned this in fact

src/screens/Portfolio/AnimatedTopBar.js
juan-cortes: nice

src/screens/Portfolio/AnimatedTopBar.js
juan-cortes: bad naming but the thing is, we animate the opacity of the bar. However, the bar while invisible is still capturing all the touch events. This flag will kill the pointer events while opacity is 0, using the offset. Renaming it to "ignorePointerEvents" would make more sense

src/screens/Distribution/index.js
juan-cortes: Nice nice

src/components/CurrencyIcon.js
juan-cortes: There's no way I can get it to work with `Currency`, `CryptoCurrency` with flowfixme yes

gre:
@juan-cortes there are some remaining comments btw

src/components/CurrencyIcon.js
gre: k we'll fix later

src/screens/Distribution/RingChart.js
juan-cortes: It was referring to a //TODO, done!

src/screens/Distribution/RingChart.js
juan-cortes: it's a vestigial code left from when I was grouping (remember the red portion that grouped less than 10% assets?) Removed

src/screens/Distribution/RingChart.js
juan-cortes: oki

gre:
@Arnaud97234 if you want to quick check all is fine after further rework

gre:
@juan-cortes can you connect the pull to refresh to the SyncBridge thing?

juan-cortes:
Connected with pull to refresh now @gre @Arnaud97234

juan-cortes:
Make sure the screen is closable if you try this, having some glitch on my local interactive version

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+import { PanResponder, View } from "react-native";
+import * as d3shape from "d3-shape";
+import Svg, { Path, G, Circle } from "react-native-svg";
+import colors from "../../colors";
+import type { DistributionItem } from "./DistributionCard";
+
+type Props = {
+  data: Array<DistributionItem>,
+  side: number,
+  style?: *,
```
gre: not used on the outside, i would remove `style`

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+  selectedKey: number => void,
+  highlight?: number,
+};
+
+class RingChart extends PureComponent<Props> {
+  static defaultProps = {
+    data: [],
```
gre: no need for default i think (empty data would have to be avoided in the first place too)

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+
+      onPanResponderMove: event => {
+        const x = event.nativeEvent.pageX - this.offsetX;
+        const y = event.nativeEvent.pageY - this.offsetY;
+
+        const y0 = side / 2;
+        const x0 = side / 2;
+        let rad = Math.atan2(y - y0, x - x0);
+        rad =
+          ((rad > 0 ? rad : 2 * Math.PI + rad) + 1.5707963268) % (2 * Math.PI);
```
gre:
```suggestion
          (2.5 * Math.PI + rad) % (2 * Math.PI);
```

to confirm it works but i think it's equivalent code

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+import * as d3shape from "d3-shape";
+import Svg, { Path, G, Circle } from "react-native-svg";
+import colors from "../../colors";
+import type { DistributionItem } from "./DistributionCard";
+
+type Props = {
+  data: Array<DistributionItem>,
+  side: number,
+  style?: *,
+  selectedKey: number => void,
```
gre:
```suggestion
  onHighlightChange: number => void,
```

probably better naming

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+  offsetY = 0;
+  highlight = -1;
+  paths: * = {};
+
+  constructor(props: Props) {
+    super(props);
+    const { highlight, side, selectedKey, data } = this.props;
+
+    if (highlight) {
+      this.highlight = highlight;
```
gre: what if `highlight=0` , isn't it index 0? so probably can remove the if?

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+          }
+        }
+        if (this.highlight !== highlight) {
+          selectedKey(highlight);
+        }
+      },
+
+      onShouldBlockNativeResponder: () => false,
+    });
+    this.paths = data.reduce(this.reducer, {
```
gre: this shouldn't be done in the constructor.

strictly speaking, it is a derivate data of the data props so it normally should live in `render()` method itself (what is data changes).

As you need this also to be used in constructor, what you can do is put that `this.paths=` logic in a method, and call if from constructor AND in `componentDidUpdate`

src/screens/Distribution/RingChart.js
```diff
@@ -0,0 +1,143 @@
...
+    };
+  };
+
+  render() {
+    const { style, highlight } = this.props;
+    return (
+      <View style={style} {...this.panResponder.panHandlers}>
+        <Svg width="100%" height="100%" viewBox="0 0 100 100">
+          <G transform="translate(50, 50)">
+            {this.paths &&
```
gre: paths is never falsy, can remove the check

src/screens/Distribution/index.js
```diff
@@ -0,0 +1,239 @@
...
+        item={item}
+        selectedKey={this.selectedKey}
+        highlighting={index === this.state.highlight}
+      />
+    </TouchableOpacity>
+  );
+
+  selectedKey = index => {
+    this.setState({ highlight: index });
+    if (typeof index === "number") {
```
gre: so this is weird too. `index` is always a number it seems in your latest impl. but it can be -1 ?

src/screens/Distribution/index.js
```diff
@@ -0,0 +1,239 @@
...
+  ListHeaderComponent = () => {
+    const { counterValueCurrency, distribution } = this.props;
+    const { highlight } = this.state;
+
+    return (
+      <View>
+        <View style={styles.header}>
+          <View style={styles.chartWrapper}>
+            <RingChart
+              side={180}
```
gre: what is this `side` number? is it the width of the thing? but it's not used as a width only in the data calculation?

if it is, probably rename to size and use it in the `style` of SVG.

it should should be calculated from the width of the screen to be responsive. (check what simulator width you were having and set like `Math.floor(deviceWidth * 180 / your_reference_size)`)

src/screens/Distribution/RingChart.js
juan-cortes: sure

src/screens/Distribution/RingChart.js
juan-cortes: sure

src/screens/Distribution/RingChart.js
juan-cortes: Seems to work 🤷‍♂

src/screens/Distribution/RingChart.js
juan-cortes: ok

PR #922, opened 2019-05-31
setup fastlane

valpinkman:
setup automation with fastlane

DONE: 

- [x] ios
	- [x] build the app
		- [x] release mode
		- [x] mock mode
		- [x] check which env is used
		- [x] use package.json version
		- [x] increment build number
		- [x] sign the build
	- [x] deploy
		- [x] testflight
		- [x] waldo
- [x] android
	- [x] build the app
		- [x] release mode
		- [x] mock mode
		- [x] check which env is used
		- [x] use package.json version
		- [x] increment version code
		- [x] sign the build
	- [x] deploy
		- [x] google play store internal
		- [x] waldo

### Type

Chore

### Context

LL-216 
LL-217

PR #925, opened 2019-06-07
LL-1406 64bit support for Android

MortalKastor:
Adding 64bit binaries for both ARM and x86, as [now required by the Google Play Store](https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html).

#### Caveat

Building a universal APK (as we're currently doing) with the embedded 64bit native libs resulted in a broken app on 32bit platforms, as it tries to load jinni 64bit libc++.

I have a gut feeling the confusion comes from this `jsc-android` requirement we had to add to `android/app/build.gradle`
````
// As required by https://github.com/react-native-community/jsc-android-buildscripts#for-react-native-version-059
packagingOptions {
    pickFirst '**/libjsc.so'
    pickFirst '**/libc++_shared.so'
}
````

Anyway, doing separate apks per ABI fixes the problem, and, as a bonus, takes less space on our users phones.
I just had to update the way we install the staging/mock apks to auto-select the relevant file for the target device.

### Type

Feature(?)

### Context

- https://ledgerhq.atlassian.net/browse/LL-1406
- LedgerHQ/lib-ledger-core#233
- LedgerHQ/lib-ledger-core-react-native-bindings#22

### Parts of the app affected

The whole Android app.

### Test plan

- Get Android system images for each of the cpu architectures with `sdkmanager 'system-images;android-25;google_apis;arm64-v8a' 'system-images;android-25;google_apis;armeabi-v7a' 'system-images;android-25;google_apis;x86' 'system-images;android-25;google_apis;x86_64'`
- Create one Android Virtual Device per cpu arch like so:
```
avdmanager create avd --name nougat-arm --abi google_apis/armeabi-v7a --package 'system-images;android-25;google_apis;armeabi-v7a' -d "Nexus 4"
avdmanager create avd --name nougat-arm64 --abi google_apis/arm64-v8a --package 'system-images;android-25;google_apis;arm64-v8a' -d "Nexus 4"
avdmanager create avd --name nougat-x86 --abi google_apis/x86 --package 'system-images;android-25;google_apis;x86' -d "Nexus 4"
avdmanager create avd --name nougat-x86_64 --abi google_apis/x86_64 --package 'system-images;android-25;google_apis;x86_64' -d "Nexus 4"
```
- Test the app in each one, but arm64 you should probably test on a real device to speed things up (emulating arm on your x86 computer is slooooooow)

### Test results

I loaded the app, imported some accounts, did a sync, and navigated a bit on each version and everything went fine.
I wasn't able to test neither BLE or USB.


MortalKastor:
It should only affect the upload of the artifacts to Waldo and the Play Store, to reflect the several apk files.
Everything else works the same (eg the build script is ran once to generate the 4 apks).

All the still relevant parts of #337 have been merged into #922 (the PR title is out of date and misleading) in a pair programming session with @valpinkman FYI

PR #931, opened 2019-06-19
Bootstrap ERC20

gre:
## ERC20 mobile work in progress

All tasks are waiting for a design, but almost everything is working. Some feature are also obviously missing. Here is non exhaustive list of big parts that remains. In bold are the biggest ones involving at least a day of work.

- [x] icon of an ERC20 Account. @IAmMorrow : you can port from your ParentCurrencyIcon desktop (need confirmation from design).
- [x] **Accounts list search**. to include token accounts. We probably will have same behavior of desktop account search (greyed part, etc..)
- [x] **Accounts screen** 
  - to show you the sub token accounts
  - Accounts screen should have the 4 new ordering option like we did on Desktop + make sure to use the same account sorting function.
- [x] **Account screen** to integrate token list.
- [ ] new **Rates settings** to implement.
- [ ] distribution: figure out the color. I have no clue where that blue of erc20 come from. was supposed to take same as eth color?
- [ ] Operation detail of a ERC20 wrongly show you fees in the token currency. should be the mainAccount currency.
- [ ] Receive: it's very likely some stuff will slightly change (e.g. we still say "Address for account" and "Ethereum address")
- [ ] Send
  - [x] editing the fees is NOT working. might have break something (it's probably just the parentId that is not propagated back from individual screens), need also to check all other custom fields (including XRP and BTC ones)
  - [x] probably some stuff will changes,.. in summary?
  - [ ] Validation step: Fees are not properly calculated for ERC20. requires rework of live-common bridge.
  - [x] When trying to send ERC20, something fails (staying on Validation screen at the moment but it correctly reached the device)
- [ ] migration: how will we inform the need to import again the ETH accounts.
- [x] Send Max: @juan-cortes can you merge your stuff back in and try to connect it?

problems that are common to desktop:

- we don't have any check on amount above balance (waiting for libcore correct balance)
- some other bugs with libcore that are tracked in JIRA


gre:
mmh, I consider these two problems as new feature that we need to do in the future.
both have their own jira tasks already so we can be good to merge? (except the lil' CI issue)

gre:
I have no clue why Flow is failing here.

meriadec:
Had same problem on vault-front few days ago, after a dependency bump.
Resolved after few iterations by adding

```
.*/core-js/.*
```

to `.flowconfig` :man_shrugging: 

for reference: https://github.com/LedgerHQ/ledger-vault-front/pull/872#issuecomment-507724271

Hope this will help :laughing:

Arnaud97234:
See remaining bugs in:
- https://ledgerhq.atlassian.net/browse/LL-1692
- https://ledgerhq.atlassian.net/browse/LL-1693
- https://ledgerhq.atlassian.net/browse/LL-1694

gre:
@Arnaud97234 it should be all good now!

src/locales/en/common.json
```diff
@@ -564,6 +564,11 @@
     "viewInExplorer": "View in explorer",
     "sending": "Sending...",
     "receiving": "Receiving...",
+    "tokenOperations": "Token transactions",
```
gre: => Token operations

src/locales/en/common.json
```diff
@@ -564,6 +564,11 @@
...
+    "tokenModal": {
+      "desc": "This operation is related to the following token transactions"
```
gre: token operations!

(disclaimer: token transaction does not exist)

src/locales/en/common.json
```diff
@@ -577,6 +582,11 @@
     "selectAll": "Select all"
   },
   "account": {
+    "tokens": {
+      "contractAddress": "Contract address",
+      "viewInExplorer": "View in explorer",
+      "howTo": "To add tokens accounts, you need to <0>receive funds</0> through your <1>Ethereum address</1>."
```
gre: @dasilvarosa could you check these? i'm not sure about the wording. `through your Ethereum address`

gre:
@IAmMorrow as soon as we hear back from @dasilvarosa can you address whatever wording feedback we get? also need to merge develop.

thanks!

PR #940, opened 2019-06-27
Ll 1525 visual bug portfolio header

IAmMorrow:
fixing a small visual bug on the portfolio screen causing the distribution button to be pushed off-screen when the content block (either error or greeting title) is too large.

### Type

bug fix

### Context

LL-1525

portfolio screen header

### DIFF
--- a/src/screens/Portfolio/Header.js
+++ b/src/screens/Portfolio/Header.js
@@ -97,6 +97,7 @@ const styles = StyleSheet.create({
   },
   content: {
     flexGrow: 1,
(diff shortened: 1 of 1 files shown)


src/screens/Portfolio/Header.js
```diff
@@ -97,6 +97,7 @@ const styles = StyleSheet.create({
   },
   content: {
     flexGrow: 1,
+    flexShrink: 1
```
IAmMorrow: easy win

PR #945, opened 2019-07-01
LL-1552 Show transaction status instead of asset icon in portfolio

juan-cortes:
Screenshot 2019-07-01 at 23 47 34
We keep the use of the account name instead of "sent/receive" but lose the currency icon.

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1552

On the portfolio history, the list rows should no longer display the currency icon but rather the transaction status indicator.


gre:
@KhalilBellakrid  is this what we want or do we also want to have the account name text be replaced by Received/Sent ?

juan-cortes:
> @khalilbenihoud is this what we want or do we also want to have the account name text be replaced by Received/Sent ?

If we show received sent here too then theres no way of knowing from which account it comes, this is the balance, we get the direction and the source 🤷‍♂

Issue #589, opened 2018-12-04
About the Android screen transitions


juan-cortes:
### Rant disclaimer

> ~All the gifs had to be made using a super bad palette because otherwise, the shadows would be invisible. It would all be white and all I'm trying to explain will be lost in words.~ 
(Updated all the gifs thanks to @meriadec )

This all started when @gre mentioned that on his device the checkboxes rendered after a while. I found out it was directly linked to the transition used by react-navigation and we changed it for something else. However, in doing so, we imposed an animation that _to my eyes_ doesn't feel natural.

For example, on onboarding, we are showing the steps coming from top to bottom and on iOS we are showing the left to right transition. This is very version dependent but if we want to make it look like the latest version available it would need to follow a transition similar to these:

http://androidxref.com/9.0.0_r3/xref/frameworks/base/core/res/res/anim/activity_open_enter.xml
http://androidxref.com/9.0.0_r3/xref/frameworks/base/core/res/res/anim/activity_close_exit.xml

Which clearly is not a slide from the bottom like I was naively doing. Implementing this in the `transitionConfig` has the added complexity of not really allowing (that I could find) the use of different interpolators for the different changes and the lack of an equivalent to `cliprect`.

 ### Why do you hate me react?
A simple fadeIn would be less glitchy and aggressive than the current version but, of course, this wasn't going to be as easy as it could be. My initial testings suggest that we keep having behavior from god knows how many years ago where the second activity showed over the first one (activities... screens in React) will show over the first one showing some lines as darker. Especially visible since we have very faint shadows. The solution for this, in case it can come handy in the future was to have borders without translucency, that way overlapping them (especially for headers and such) avoided this aberration.

![preshadow](https://user-images.githubusercontent.com/4631227/49437950-2e273080-f7bd-11e8-8a07-85acf04fa881.gif)

Not even sure it's the exact same behavior or something completely unrelated, but of course this looks terrible, I decided to remove the `elevation` we have for the navigation headers and replace if with a bottom border with a solid color. Note the difference on that header part of the navigation.
![postshadow](https://user-images.githubusercontent.com/4631227/49437962-341d1180-f7bd-11e8-8e08-460183a8afcd.gif)

### Focusing on the translation

Limiting the transition to only a y-axis translation made it better than nothing, but still a bit abrupt, since it simply disappears after it moves down. The offset of around 50 units comes from the above animations. 
[...]

PR #948, opened 2019-07-04
LL-1392  Sendmax functionality

juan-cortes:
In the current state of common, it only works with bitcoin family since the ethereum bridge doesn't implement a way of setting the `useAllAmount` extra (~i think this is the cause~ it was). Once https://github.com/LedgerHQ/ledger-live-common/pull/296 is merged in common and we bump it, this pr would work for eth too.

Group 2

### Type

Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-1392

Send flow

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -845,7 +845,8 @@
       "input": "Enter address"
     },
     "amount": {
-      "available": "<1><0>{{amount}}</></> available",
+      "available": "Total available",
+      "useMax": "Use max",
       "loadingNetwork": "Loading Network fees...",
       "noRateProvider": "Not available"
     },
--- a/src/screens/SendFunds/03-Amount.js
+++ b/src/screens/SendFunds/03-Amount.js
@@ -5,6 +5,7 @@ import {
   StyleSheet,
   TouchableWithoutFeedback,
   Keyboard,
+  Switch,
 } from "react-native";
 import { SafeAreaView } from "react-navigation";
 import { connect } from "react-redux";
@@ -51,6 +52,7 @@ type State = {
   txValidationWarning: ?Error,
   totalSpent: ?BigNumber,
   leaving: boolean,
+  maxAmount: ?BigNumber,
 };
 
 const similarError = (a, b) =>
@@ -80,6 +82,7 @@ class SendAmount extends Component<Props, State> {
       txValidationWarning: null,
       totalSpent: null,
       leaving: false,
+      maxAmount: null,
     };
   }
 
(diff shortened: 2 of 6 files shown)


src/locales/en/common.json
```diff
@@ -845,7 +845,8 @@
       "input": "Enter address"
     },
     "amount": {
-      "available": "<1><0>{{amount}}</></> available",
+      "available": "Total available",
+      "useMax": "Use max",
```
gre: @dasilvarosa can you confirm wording? thanks!

src/screens/SendFunds/03-Amount.js
```diff
@@ -80,6 +83,8 @@ class SendAmount extends Component<Props, State> {
       txValidationWarning: null,
       totalSpent: null,
       leaving: false,
+      useAllAmount: false,
```
gre: unless i miss something, this shouldn't be a state because it's already part of `transaction`

the same way we don't have `amount` in state but we use this in render():   `const amount = bridge.getTransactionAmount(mainAccount, transaction);`

src/screens/SendFunds/03-Amount.js
```diff
@@ -311,27 +355,39 @@ class SendAmount extends Component<Props, State> {
...
-                      {"text"}
-                    </Trans>
-                  </LText>
+                    </View>
+                    <View style={styles.availableRight}>
+                      <LText style={styles.maxLabel}>
+                        <Trans i18nKey="send.amount.useMax" />
+                      </LText>
+                      <Switch
+                        style={{ opacity: 0.99 }}
```
gre: why? if there is a reason, please comment a `// FIXME` explanation and/or link (typically we later can check for these fixme comment)

src/screens/SendFunds/04-Summary.js
```diff
@@ -159,6 +163,35 @@ class SendSummary extends Component<
...
+      transaction,
+      "useAllAmount",
+    );
+
+    if (nonce !== this.nonceMaxAmount) return;
+
+    let maxAmount;
+    if (useAllAmount)
+      maxAmount = await bridge.getMaxAmount(mainAccount, transaction);
+
```
gre: a `if (nonce !== this.nonceMaxAmount) return;` needs to be here (typically after every await)

src/screens/SendFunds/03-Amount.js
```diff
@@ -213,6 +218,42 @@ class SendAmount extends Component<Props, State> {
...
+      transaction,
+      BigNumber(0),
+    );
+
+    if (nonce !== this.nonceUseAllAmount) return;
+
+    let maxAmount;
+    if (useAllAmount)
+      maxAmount = await bridge.getMaxAmount(mainAccount, transaction);
+
```
gre: same as commented below

src/screens/SendFunds/04-Summary.js
```diff
@@ -205,7 +238,7 @@ class SendSummary extends Component<
           <SummaryAmountSection
             account={account}
             parentAccount={parentAccount}
-            amount={amount}
+            amount={maxAmount || amount}
```
gre: isn't maxAmount always set? would it always display amount as a max in that code?

src/screens/SendFunds/03-Amount.js
juan-cortes: Any animation involving opacity makes some buttons like radio buttons and switches flash on some versions of Android, this was the solution I ranted about at https://github.com/LedgerHQ/ledger-live-mobile/issues/589 don't know if that's detailed enough for you, I've been using that ever since. If you think there's a need to document this behaviour further, I can add a reference to the link above each time we use it

src/screens/SendFunds/03-Amount.js
juan-cortes: makes sense

src/screens/SendFunds/04-Summary.js
juan-cortes: got it

juan-cortes:
Everything except the wording check and the 0.99 comment updated, I'll put another pr for the comment to not mix things up, so if @dasilvarosa approves the wording, i'm done with this pr (🤞  @Arnaud97234 )

gre:
live-common@7.1.0

gre:
@IAmMorrow will test it a bit. at worse, we'll re-review wording at topic/erc20 merge time

PR #950, opened 2019-07-04
slider for Ethereum transaction

valpinkman:
Adds a slider to select the gas of a transaction for Ethereum-like currencies

### Type

Feat

### Context

LL-1363

Send Flow => step 4 Summary => edit gas

IMG_9701E8CB6E84-1

### DIFF
--- a/src/families/ethereum/EditFeeUnitEthereum.js
+++ b/src/families/ethereum/EditFeeUnitEthereum.js
@@ -1,0 +1,198 @@
+// @flow
+import React, { useMemo, useState, useCallback } from "react";
+import { View, StyleSheet } from "react-native";
+import { translate } from "react-i18next";
+import Slider from "react-native-slider";
+import type { NavigationScreenProp } from "react-navigation";
+import type { Account, TokenAccount } from "@ledgerhq/live-common/lib/types";
+import type { Transaction } from "@ledgerhq/live-common/lib/bridge/EthereumJSBridge";
+import {
+  inferDynamicRange,
+  reverseRangeIndex,
+  projectRangeIndex,
+} from "@ledgerhq/live-common/lib/range";
+import { getAccountBridge } from "@ledgerhq/live-common/lib/bridge";
+import { getMainAccount } from "@ledgerhq/live-common/lib/account";
+import { BigNumber } from "bignumber.js";
+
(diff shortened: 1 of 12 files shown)


src/families/ethereum/EditFeeUnitEthereum.js
```diff
@@ -0,0 +1,202 @@
...
+      onValueChange={setValueIndex}
+      minimumValue={0}
+      maximumValue={range.steps - 1}
+      thumbTintColor={colors.live}
+      minimumTrackTintColor={colors.live}
+    />
+  );
+});
+
+const fallbackGasPrice = BigNumber(10000000000);
```
gre: probably not needed

src/families/ethereum/EditFeeUnitEthereum.js
```diff
@@ -0,0 +1,202 @@
...
+  const [gasPrice, setGasPrice] = useState(() =>
+    bridge.getTransactionExtra(mainAccount, transaction, "gasPrice"),
+  );
+  const feeCustomUnit = bridge.getTransactionExtra(
+    mainAccount,
+    transaction,
+    "feeCustomUnit",
+  );
+
+  const fees = bridge.getTransactionNetworkInfo(mainAccount, transaction);
```
gre: this is guaranteed to be filled in mobile paradigm. You can assume it's set. so feel free to `if (!fees) return null` after this line so you don't need a `fallbackGasPrice`

src/locales/en/common.json
```diff
@@ -878,6 +880,8 @@
       "title": "Network fees",
       "validate": "Confirm",
       "required": "Fees are required",
+      "chooseGas": "Choose a gas price",
+      "higherFaster": "Higher gas price means a faster confirmation",
```
gre: @dasilvarosa please confirm. thanks! (maybe you already did on slack, just to be sure :p )

valpinkman: on slack it seems that the wording did come from @dasilvarosa

gre:
to be tested that all custom summary edit screen still works in all of the 3 cases: ETH, BTC, XRP (edit fee, edit tag for xrp)

valpinkman:
I did test for BTC (did not have any XRP) and it seems to still be okay

PR #952, opened 2019-07-08
fixed distribution placeholders when no countervalue

IAmMorrow:
Fixed distribution flatlist item keys were not generated
Hiding distribution asset card percent and countervalue when no countervalue is available

### Type

Bug Fix

Distribution Screen

### DIFF
--- a/src/screens/Distribution/index.js
+++ b/src/screens/Distribution/index.js
@@ -76,7 +76,7 @@ class Distribution extends PureComponent<Props, *> {
     this.flatListRef.scrollToIndex({ index }, true);
   };
 
-  keyExtractor = item => item.id;
+  keyExtractor = item => item.currency.id;
 
   ListHeaderComponent = () => {
     const { counterValueCurrency, distribution } = this.props;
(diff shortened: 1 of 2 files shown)


src/screens/Distribution/index.js
```diff
@@ -76,7 +76,7 @@ class Distribution extends PureComponent<Props, *> {
     this.flatListRef.scrollToIndex({ index }, true);
   };
 
-  keyExtractor = item => item.id;
+  keyExtractor = (item, index) => index;
```
gre: minor feedback: item.?currency?.id

IAmMorrow: agreed

src/screens/Distribution/index.js
```diff
@@ -127,6 +127,7 @@ class Distribution extends PureComponent<Props, *> {
   render() {
     const { distribution } = this.props;
 
+    console.log(distribution.list)
```
gre: oops

IAmMorrow: 😭

PR #955, opened 2019-07-09
added parent currency icons

IAmMorrow:
image

Add double icons (parent && token) to the account header and distribution panels

### Type

Ui Polish

Distribution screen / Token Account Screen


gre:
nice

PR #966, opened 2019-07-18
updated pre-selecting account strategy

IAmMorrow:
if no existing account is found, will pre-select the last potential new account in the list.

### Type

UI Polish

### Context

LL-1673

Add account -> Scan account screen

### DIFF
--- a/src/screens/AddAccounts/03-Accounts.js
+++ b/src/screens/AddAccounts/03-Accounts.js
@@ -7,6 +7,7 @@ import i18next from "i18next";
 import { isAccountEmpty } from "@ledgerhq/live-common/lib/account";
 import { createStructuredSelector } from "reselect";
 import uniq from "lodash/uniq";
+import findLast from "lodash/findLast";
 import { translate, Trans } from "react-i18next";
 import { StyleSheet, View } from "react-native";
 // $FlowFixMe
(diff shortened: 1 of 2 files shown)


src/screens/AddAccounts/03-Accounts.js
```diff
@@ -7,6 +7,7 @@ import i18next from "i18next";
 import { isAccountEmpty } from "@ledgerhq/live-common/lib/account";
 import { createStructuredSelector } from "reselect";
 import uniq from "lodash/uniq";
+import reduce from "lodash/reduce";
```
gre: reduce is part of Array. not need to use

src/screens/AddAccounts/03-Accounts.js
```diff
@@ -99,19 +100,47 @@ class AddAccountsAccounts extends PureComponent<Props, State> {
...
-            if (!isAccountEmpty(account) && !this.isExistingAccount(account)) {
-              // $FlowFixMe
-              patch.selectedIds = [...selectedIds, account.id];
-            }
+            // $FlowFixMe
+            patch.selectedIds = onlyNewAccounts
+              ? [account.id]
+              : reduce(
+                  scanRecords,
+                  (acc, record) => {
```
gre: it seems like you just want `.filter `anyway?

PR #970, opened 2019-07-23
LL-1693 Fixes send max broken button for erc20

juan-cortes:
The sendmax button was not working for tokens because the bridge was not being retrieved correctly.

### Type

BugFix

### Context

https://ledgerhq.atlassian.net/projects/LL/issues/LL-1693

Send flow, amount step, sending from token account


juan-cortes:
XRP shouldn't have send max at all, I'll check. 
Please look at this only from the erc20 perspective @Arnaud97234

gre:
pretty straightforward to merge

PR #974, opened 2019-07-23
LL-1633: Add LiveQR export on mobile

pbondoer:
This PR adds the LiveQR export functionality to mobile to be able to import from another mobile device.

For now it's hidden in the :sparkles: secret debug menu :sparkles: but it will be more user-facing #soon

### :movie_camera: Video demo

[Click here](
https://www.youtube.com/watch?v=oUVsdabfl84)

https://www.youtube.com/watch?v=oUVsdabfl84

### :ok_hand: Type

Feature

### :mag: Context

LL-1633

### :clipboard: Parts of the app affected / Test plan

* Debug menu > Export accounts
* Import accounts


src/reducers/settings.js
```diff
@@ -323,4 +323,15 @@ export const countervalueFirstSelector = (state: State) =>
...
   Platform.OS !== "android" && state.settings.readOnlyModeEnabled;
 
+// $FlowFixMe
+export const exportSettingsSelector = createSelector(
+  counterValueCurrencySelector,
+  state => state.settings.currenciesSettings,
+  (counterValueCurrency, currenciesSettings) => ({
+    counterValue: counterValueCurrency.ticker,
+    currenciesSettings,
+    developerModeEnabled: __DEV__,
```
gre: mmh, it's more the actual value of the settings developerModeEnabled. (this enable or not the testnet coins)

gre: also, the signature of that liveqr settings have recently changed https://github.com/LedgerHQ/ledger-live-common/blob/master/src/cross.js (in recent live-common version)
and it is already the case on desktop : https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/reducers/settings.js#L248-L259

depending on which version of live-common we are on, we might want to HODL this PR for this new live-common version and adapt the present code.

pbondoer: Couldn't understand how `pairExchanges` was stored on mobile, and for dev mode I couldn't find the switch in the settings reducer either. I'll look again tomorow :)

pbondoer: Couldn't understand how `pairExchanges` was stored on mobile, and for dev mode I couldn't find the switch in the settings reducer either. I'll look again tomorow :)

pbondoer:
After internal discussion we'll wait for `topic/erc20` to merge :ok_hand:

gre:
@pbondoer it's low in this current sprint but you should later be able to rebase develop and have latest stuff that was blocking this work :)

pbondoer:
@gre @Arnaud97234 Since we merged `topic/erc20` I have pushed some changes, could you please review again? :mag:

src/reducers/settings.js
```diff
@@ -85,6 +86,7 @@ const INITIAL_STATE: SettingsState = {
   readOnlyModeEnabled: !Config.DISABLE_READ_ONLY,
   experimentalUSBEnabled: false,
   countervalueFirst: false,
+  developerMode: !!__DEV__,
```
gre: mmh I don't think this is what we want. the developerModeEnabled is actually mapped to `getEnv("DEV_MODE")`, which is **not** the same as __DEV__. it's basically a user's settings that enable or not the testnet apps and currencies.

gre: (it's ok to call getEnv("DEV_MODE") at runtime from the selector because it's going to be set when you enter the screen)

pbondoer: @gre This is what we do on desktop though... https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/reducers/settings.js#L88

gre: I think that `developerMode` is a dead code unfortunately! seems like it was supposed to get dropped as we move it all to the DEV_MODE env. it's used in bunch of place in live-common and is the way we filter apps and currencies.

gre: the question is do we restore it back correctly, i think we don't. i wonder if we want to really use that in the "export" if it's not really working. We might not want to export the experimental settings. 

wdyt @aramab  ?

pbondoer: Export is definitely using that, I followed the selector there, so if there's changes to make there I'll port them over to desktop too :)

aramab: I would export it IMO

PR #982, opened 2019-07-30
LL-1709 removing fees on IN operations

IAmMorrow:
Removing the fee column from IN operations (receives)

### Type

UI Polish

### Context

LL-1709

Operation Data Screen -> Fee row

### DIFF
--- a/src/screens/OperationDetails/Content.js
+++ b/src/screens/OperationDetails/Content.js
@@ -224,35 +224,41 @@ class Content extends PureComponent<Props, State> {
             })}
           </LText>
         </View>
-        <View style={styles.section}>
-          <LText style={styles.sectionTitle}>
-            <Trans i18nKey="operationDetails.fees" />
-          </LText>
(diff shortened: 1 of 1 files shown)


src/screens/OperationDetails/Content.js
```diff
@@ -224,35 +224,41 @@ class Content extends PureComponent<Props, State> {
             })}
           </LText>
         </View>
-        <View style={styles.section}>
-          <LText style={styles.sectionTitle}>
-            <Trans i18nKey="operationDetails.fees" />
-          </LText>
-          {operation.fee ? (
-            <View style={styles.feeValueContainer}>
+        {operation.type !== "IN" ? (
```
gre: prefer `==="OUT"`

"NONE" case is not rendered, but if it was, we would not want fees neither, it's really just the OUT

IAmMorrow: Done !

PR #965, opened 2019-07-17
bump live-common / libcore. internal transactions. import accounts from desktop to dedup/migrate

gre:
Rework the import from accounts logic in order to also migrate and dedup the accounts when LIBCORE_EXPERIMENTAL is on and when old ethereumjs account are on mobile.

What is expected here?

- Exporting from desktop works with libcore accounts too but if mobile don't enable yet the experimental, it will make the experimental ethereum libcore appearing under unsupported. It should also properly migrate old eth accounts.
- there will no longer be dup cases. (fixes LL-1716 LL-1657)
- internal transactions

PR #983, opened 2019-07-30
LL-1477 Hide empty token accounts

MortalKastor:
Adds a toggle in the _General_ settings page to hide token accounts with 0 balance.

This selectively hides the accounts only in some part of the UI:

- Accounts screen
- Token list in parent account screen
- Send modal (including search)

Other parts shouldn't be affected by this toggle, like (non exhaustive list):
- Operations
- Operation details
- Receive flow
- …

HODL for now until #965 gets merged

### Type

Feature

### Context

LedgerHQ/ledger-live-common#294
LedgerHQ/ledger-live-desktop#2198
https://ledgerhq.atlassian.net/browse/LL-1477

### Parts of app affected

- Accounts screen
- Token list in parent account screen
- Send modal (including search)

### DIFF
--- a/src/actions/settings.js
+++ b/src/actions/settings.js
@@ -1,6 +1,5 @@
 // @flow
 
-import { setEnvUnsafe } from "@ledgerhq/live-common/lib/env";
 import type { Currency } from "@ledgerhq/live-common/lib/types";
 
 export type CurrencySettings = {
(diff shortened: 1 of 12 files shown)


src/actions/settings.js
```diff
@@ -106,3 +107,14 @@ export const installAppFirstTime = () => ({
 export const switchCountervalueFirst = () => ({
   type: "SETTINGS_SWITCH_COUNTERVALUE_FIRST",
 });
+
+export const setHideEmptyTokenAccounts = (hideEmptyTokenAccounts: boolean) => (
+  dispatch: *,
+) => {
+  if (setEnvUnsafe("HIDE_EMPTY_TOKEN_ACCOUNTS", hideEmptyTokenAccounts)) {
+    dispatch({
+      type: "SETTINGS_HIDE_EMPTY_TOKEN_ACCOUNTS",
```
gre: mmh I think it's an antipattern to have an env both existing in live-common's env and in store. ideally we need to keep state in one place.

if you need something to update based on that env, typically for the checkbox, just use `useEnv` (check with @valpinkman who wrote them. i'm not remembering if they are in live-common yet)

MortalKastor: >  I think it's an antipattern to have an env both existing in live-common's env and in store

The state of the env has to be saved for it to be restored after the app has been closed. For the desktop implementation I found it more fitting to save it with all the other settings in the redux store rather than follow what has been done for experimentals which are saved in localStorage. I just did it in the same fashion here.

> If you need something to update based on that env, typically for the checkbox, just use `useEnv`

I'm completely at fault on this point, that was pretty stupid of me to rely on the setting state to determine wether to update components while I could have subscribed in a way or another to the env state itself 😅
`useEnv` is not in live-common yet, but `withEnv` can do the trick

gre:
@valpinkman could you take this over? to solve some minor code conflicts between recent @IAmMorrow commit vs this PR.

thx

valpinkman:
@gre done

gre:
@valpinkman side note for future: a merge commit is preferred in this situation so you don't rewrite history ;)

(i'm going to squash and merge anyway because it's not an atomic commit ;))

valpinkman:
@gre I will do that next time then ^^

PR #984, opened 2019-07-30
LL-1708 Allow user to go back to operation details from suboperation

juan-cortes:
Overrides the left/right buttons on operation details if we are in a suboperation allowing the user to go back to source operation without closing the whole thing.

goback

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1708

### Parts of the app aff

ected / Test plan

Operation details for an operation with suboperations

### DIFF
--- a/src/components/OperationRow.js
+++ b/src/components/OperationRow.js
@@ -32,6 +32,7 @@ type Props = {
   navigation: *,
   multipleAccounts?: boolean,
   isLast: boolean,
+  isSubOperation?: boolean,
 };
 
 const placeholderProps = {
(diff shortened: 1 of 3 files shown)


src/components/OperationRow.js
```diff
@@ -32,6 +32,7 @@ type Props = {
   navigation: *,
   multipleAccounts?: boolean,
   isLast: boolean,
+  subOperation?: boolean,
```
gre: minor feedback on the naming, i suggest as it's a bool, to not confuse with `Operation` to call it `isSubOperation` or maybe `withBackButton` ? 🤔

juan-cortes: Agreed, rebased, `isSubOperation` for now, I'm not rebasing again! :trollface:

PR #985, opened 2019-07-31
LL-1712 adapting font size to width to display more

IAmMorrow:
Currency input will now adjust the font size regarding of the available width against the number of characters in order to display the largest string possible. The font is then dynamically adjusted.

image

### Type

UI Polish

### Context

LL-1712

Send Screen -> Currency Input

### DIFF
--- a/src/components/CurrencyInput.js
+++ b/src/components/CurrencyInput.js
@@ -1,12 +1,14 @@
 // @flow
 import React, { PureComponent } from "react";
-import { TextInput, StyleSheet, View } from "react-native";
+import { TextInput, StyleSheet, View, Dimensions } from "react-native";
 import { BigNumber } from "bignumber.js";
+
 import {
   formatCurrencyUnit,
   sanitizeValueString,
 } from "@ledgerhq/live-common/lib/currencies";
 import noop from "lodash/noop";
+import clamp from "lodash/clamp";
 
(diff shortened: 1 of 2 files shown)


src/components/CurrencyInput.js
```diff
@@ -139,16 +141,27 @@ class CurrencyInput extends PureComponent<Props, State> {
...
     const { displayValue } = this.state;
+
+    // calculating an approximative font size
+    const screenWidth = Dimensions.get("window").width * 0.75;
+    const dynamicFontSize = Math.round(
+      clamp(
+        Math.sqrt((screenWidth * 32) / displayValue.length),
+        8,
+        isActive ? 32 : 24,
+      ),
```
gre: that's a bit dangerous as i doubt we use a monospace font. mmmh

IAmMorrow: Yes but sadly it works only for <Text /> :(

The font size is capped between 8 and 32 so it can't really get out of hand.

IAmMorrow: Sadly the native way only works for text components and this is actually an input ...

So another solution would be to replace inputs by texts whenever the send max is on.

In this PR's approach, the font size is capped between 8 and 32 so it cannot really get out of hand.

PR #988, opened 2019-08-01
LL-1688: Implement rates settings

pbondoer:
This PR implements rates settings per available currency pair.

### :ok_hand: Type
Feature

### :mag: Context
LL-1688

Settings > Crypto assets > Rates


src/screens/Settings/index.js
```diff
@@ -43,17 +44,8 @@ class Settings extends Component<Props, *> {
...
-      return currencies.length < 2
-        ? navigation.navigate("CurrencySettings", {
-            currencyId: currencies[0].id,
-          })
-        : navigation.navigate("CurrenciesList");
-    }
-    return navigation.navigate(screenName);
-  };
+  navigateTo = (screenName: string) =>
+    this.props.navigation.navigate(screenName);
```
gre: We can probably inline this in render

gre:
regarding your comment @Arnaud97234 ,
there are either a paradigm shift to accept or a rework to do.

**Solution 1:**

we accept that first screen to just show you the settings without calling network. That means we have what this PR shows: a list of rates with exchange ids.

**Solution 2:**

We do like Desktop and we call ALL http methods to fetch ALL the exchanges for each rates. 

Proside:
- we would be able to display the exchange name instead of the id
- we would be able to have the error case Arnaud is talking about
- You can open the "edit exchanges" screen without that screen to fetch the data as the parent would load it so it's a bit better UX (no spinner in the screen, spinner would be on main screen)

Downside:
- it's a bit more intense for the network and less mobile optimized.
- it's more work to implement this.




**My opinion:**

I think Solution 1 is shippable so we can solve the problem of not being able to configure the erc20 rates.
I think Solution 2 is a good to have in the future, for consistency with desktop and especially for the fact we don't display the exchange name but technical ids to the user.


So I would recommend we have a second iteration in next sprint @aramab

pbondoer:
@gre I think what @Arnaud97234 means is that we need to show "no exchanges" in the select screen, when we do perform a network call - and we can know this quite easily simply by checking the returned array, I just need to implement it in the generic component I think

gre:
oh yeah good point :) (but solution 2 is still something to consider in future aha :p)

Arnaud97234:
Coins to FIAT and Coin to Coin sidebar is still missing on IOS.
@gre do we want the placeholder in the Countervalue list in case of no rateprovider available.

aramab:
Placeholder : "No rate provider available for this asset"

src/screens/Settings/CryptoAssets/Rates/RatesList.js
```diff
@@ -0,0 +1,157 @@
...
+  );
+
+  // prettier-ignore
+  keyExtractor = ({ from, to }: Pair): string => `${from.name || ""}-${to.name || ""}`;
+
+  render() {
+    const { pairs } = this.props;
+
+    return (
+      <ScrollView style={styles.root}>
```
MortalKastor: You should use FlatList's `ListHeaderComponent` prop rather than nesting scrollables

PR #989, opened 2019-08-01
Uses live-common logic for addAccounts

gre:
Reuse logic of live-common for the add accounts logic that have the logic for sections, have "accounts to migrate" section and also will correctly dedup the accounts.

### Type

bug fix / feature

### Context

LL-1716

add accounts


Arnaud97234:
Behavior is correct, do we wait for the wording before merging @gre ?

gre:
wording is correct because it's the same as on desktop.

PR #999, opened 2019-08-07
added a button to expand a shortened token list

IAmMorrow:
If an eth account has funds but no token, display a touchable.

If an account has 3 tokens or less , render the full list

If an account has more than 3 tokens, render a collapsable list

If an account has no funds and no token, display a generic empty state

image

### Type

UI Polish

### Context

N / A

Account -> Token list

### DIFF
--- a/src/screens/Account/TokenAccountsList.js
+++ b/src/screens/Account/TokenAccountsList.js
@@ -1,20 +1,22 @@
 // @flow
 
-import React, { useCallback } from "react";
+import React, { useCallback, useState } from "react";
 import { Trans } from "react-i18next";
+import take from "lodash/take";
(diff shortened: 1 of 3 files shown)


src/screens/Account/TokenAccountsList.js
```diff
@@ -63,36 +71,104 @@ const Card = ({ children }: { children: any }) => (
...
+  navigation,
+  accountId,
 }: {
   tokenAccounts: TokenAccount[],
   onAccountPress: TokenAccount => *,
+  navigation: *,
+  accountId: string,
 }) => {
+  const [isCollapsed, setCollapsed] = useState(true);
+  const renderHeader = useCallback(
```
gre: post PR feedback: a dedicated Header component would be preferred in this kind of big comopnents.

PR #1002, opened 2019-08-19
upgrade live-common version

gre:
required by #1000 

Upgraded live-common with 2 past weeks work.

Check everything is still working fine.

Please see list of changes in https://github.com/LedgerHQ/ledger-live-desktop/pull/2269

PR #1003, opened 2019-08-19
Feat/insufficient gas

valpinkman:
Check for gas when sending ERC20 tokens

depends on #1002 

### Type

Feat

### Context

LL-1621

Send Flow

IMG_C2CDB9B31163-1


src/locales/en/common.json
```diff
@@ -205,6 +205,9 @@
       "title": "Oops, insufficient balance",
       "description": "Make sure the account to debit has sufficient balance"
     },
+    "NotEnoughGas": {
+      "title": "Oops, insufficient gas"
```
juan-cortes: >_ETH balance in parent account insufficient for network fees_

According to https://github.com/LedgerHQ/ledger-live-desktop/blob/63b38a673c9d3d50b7997014557f37888b49dc7f/static/i18n/es/app.json#L1056

juan-cortes: 🤷‍♂

gre: Yeah agree we need same wording. If we can change the wording, please also update the screenshot before we confirm with the team this is where we need the fix to be.

valpinkman:
IMG_4C8B4E995185-1

the text seems WWAAAAYYYY too long for me.

@gre @dekkaki @khalilbenihoud

juan-cortes:
That looks disgustingly long  👀

gre:
I think the point is more to start a discussion anyway, we're not sure what is the best solution yet but it's important we at least want to block at this point because you shouldn't be able to go farther than this if you can't spend the transaction right?

we should have an internal discussion what what would be a better UX if we want something better :)

gre:
=> probably need to move the error on the summary so we can change the gas afterwards and also handle case where you would augment the gas.

faith (approved):

LFBarreto:
the issue seems fixed now closing this

PR #1009, opened 2019-08-22
Introduce mock data preset

gre:
allow to statically starts from mock data in MOCK mode.

mock_data_preset


src/mocks/1.json
```diff
@@ -0,0 +1,2872 @@
+{
```
juan-cortes: I don't really like carrying mock data in the repo when it can be generated, but perhaps 128kb is not that much for the gain it provides. We'd have to stay on top of it so that it's valid.

juan-cortes: Dates, too

gre: good point. i'm not sure how i can also only embed it in MOCK mode.

that was just an example, it's not supposed to be generated **but we would use real libcore accounts potentially**.
we could just remove operations and set `[]` and libcore would re-fill it. so data could be minimal.

gre:
i'm going to HODL this PR until we fill up actual interesting data sets ;)

gre:
no longer needed

PR #1017, opened 2019-08-27
LL-1786: Add ERC20 empty state

pbondoer:
> :information_source: Desktop: https://github.com/LedgerHQ/ledger-live-desktop/pull/2299

This PR adds an empty state for ERC20.

### :ok_hand: Type
UX Polish

### :mag: Context
LL-1786

Ethereum accout screen
Other account screens (check ERC20 CTA isnt displayed)

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -601,6 +601,7 @@
     "emptyState": {
       "title": "No crypto assets yet?",
       "desc": "Make sure the <1><0>{{managerAppName}}</0></1> app is installed and start receiving",
+        "descToken": "Make sure the <1><0>{{managerAppName}}</0></1> app is installed and start receiving <3><0>{{currencyTicker}}</0></3> and <5><0>{{tokenType}}</0> tokens</5>",
       "buttons": {
         "receiveFunds": "Receive"
       }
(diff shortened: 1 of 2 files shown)


src/locales/en/common.json
```diff
@@ -601,6 +601,7 @@
     "emptyState": {
       "title": "No crypto assets yet?",
       "desc": "Make sure the <1><0>{{managerAppName}}</0></1> app is installed and start receiving",
+        "descToken": "Make sure the <1><0>{{managerAppName}}</0></1> app is installed and start receiving <3><0>{{currencyTicker}}</0></3> and <5>ERC20 tokens</5>",
```
juan-cortes: ERC20 tokens needs to be generalised to get the token type, which would come from `tokenType.toUpperCase()`

src/screens/Account/EmptyStateAccount.js
```diff
@@ -26,6 +26,8 @@ class EmptyStateAccount extends PureComponent<{
   render() {
     const { account, parentAccount } = this.props;
     const mainAccount = getMainAccount(account, parentAccount);
+    const isEthereum = mainAccount.currency.ticker === "ETH";
```
juan-cortes: I think, same as with the previous comment, this needs to come from whether the currency has tokens or not, which is available from `account.tokenAccounts` being undefined or not, it should be empty/filled for a libcore eth and undefined for example for btc

pbondoer: Thank you! I was looking for the right way to do this, didn't feel right at all

pbondoer:
@juan-cortes Done, needs https://github.com/LedgerHQ/ledger-live-common/pull/334, waiting for live-common bump

PR #1018, opened 2019-08-27
make sure the bottom inset is always set in safeareaview

valpinkman:
makes sure there is always enough margin on the bottom of the iphone X (and all around full screens phones)

### Type

UI Ploish

### Context

LL-1794

Everywhere


valpinkman:
@Arnaud97234 added safeareaview to 1st and 3rd screen in migrate account

PR #1032, opened 2019-09-10
LL-1848 Show architecture + version for android

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1848

The about section of settings should reflect a compound value for the Version on Android. iOS should remain the same. Also make sure I'm not breaking the stats :trollface:

### DIFF
--- a/src/logic/cleanBuildVersion.js
+++ b/src/logic/cleanBuildVersion.js
@@ -1,0 +1,34 @@
+// @flow
+
+import { Platform } from "react-native";
+
+const mega = 1048576;
+export const getAndroidArchitecture = (buildVersion?: ?string) => {
+  const buildVersionNumber = parseInt(buildVersion, 10);
+  if (!buildVersionNumber) return "";
+
+  if (Platform.OS === "android" && buildVersionNumber) {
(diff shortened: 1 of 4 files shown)


src/logic/cleanBuildVersion.js
```diff
@@ -0,0 +1,33 @@
...
+import { Platform } from "react-native";
+
+const mega = 1048576;
+export const getAndroidArchitecture = (buildVersion?: ?string) => {
+  const buildVersionNumber = parseInt(buildVersion, 10);
+  if (!buildVersionNumber) return "";
+
+  if (Platform.OS === "android" && buildVersionNumber) {
+    const platforms = ["armeabi-v7a", "x86", "arm64-v8a", "x86_64"];
+    return platforms[Math.floor(buildVersionNumber / mega) - 1];
```
gre: are we sure about the `- 1`?

gre: also could you add a comment with a link to the sourcecode that do this calculation? it's somewhere in a build gradle script

juan-cortes:
What's blocking this?

PR #1033, opened 2019-09-10
Remove references to old module, bump fork

juan-cortes:
There were apparently lingering references to the old name of the module (that's what happenes when you've never done this before) and also some references on the fork itself. I had to remove them from here and bump a new version of the fork https://github.com/LedgerHQ/react-native-touch-id

### Type

Bug Fix

Make sure you are able to build the project. I was able to build locally and test the touch id functionality.

### DIFF
--- a/android/app/src/main/java/com/ledger/live/MainApplication.java
+++ b/android/app/src/main/java/com/ledger/live/MainApplication.java
@@ -22,7 +22,6 @@ import org.reactnative.camera.RNCameraPackage;
 import io.fixd.rctlocale.RCTLocalePackage;
 import io.sentry.RNSentryPackage;
 import org.devio.rn.splashscreen.SplashScreenReactPackage;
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/com/ledger/live/MainApplication.java
```diff
@@ -55,7 +54,6 @@ public boolean getUseDeveloperSupport() {
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
           new MainReactPackage(),
-            new FingerprintAuthPackage(),
```
gre: mmh this code shouldn't be removed right?

juan-cortes: It was duplicated, it's still imported and listed in the packages in the current file.

PR #1037, opened 2019-09-11
LL-1779 Disable the import option when we are onboarding a nanoX from readonly mode

juan-cortes:
image

Disable the import from qr code if readonly mode is enabled and we are on the nano x option. This only applies to iOS since on Android we have no read only mode.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1779

### DIFF
--- a/src/screens/Onboarding/steps/get-started.js
+++ b/src/screens/Onboarding/steps/get-started.js
@@ -1,10 +1,13 @@
 // @flow
 import React, { Component, PureComponent } from "react";
 import { StyleSheet, View, Linking, Platform } from "react-native";
+import { connect } from "react-redux";
 import { Trans } from "react-i18next";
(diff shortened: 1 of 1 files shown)


juan-cortes:
What's blocking this?

juan-cortes:
@Arnaud97234 following your request import should be disabled after the initial onboarding for the nano x branch of the onboarding since iOS will have no options of nano s/blue otherwise.

juan-cortes:
I'll get back to this one after the getTransactionStatus branch, switching is a nightmare.

juan-cortes:
Forgot to get back to you @Arnaud97234 on this, but unless I'm misunderstanding you, this pr does disables the import option only for nano x regardless of it being android or iphone (from our chat) but keeps it when the selected device is not nano x.

PR #1043, opened 2019-09-18
LL-1826 Adds Asset screen

juan-cortes:
New screen, only accessible in this commit through the asset distribution screen. This will be changed in future commits to follow the new portfolio flow.

### Type

Half Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1826

New screen, accessible from the asset distribution rows.

### DIFF
--- a/src/actions/portfolio.js
+++ b/src/actions/portfolio.js
@@ -1,12 +1,16 @@
 // @flow
+import { flattenAccounts } from "@ledgerhq/live-common/lib/account";
 import {
   getBalanceHistoryWithCountervalue,
+  getCurrencyPortfolio,
   getPortfolio,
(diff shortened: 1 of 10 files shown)


juan-cortes:
So what are we gong to do with this? @gre

gre:
to me we ship it with a "dev" way to access it for now. or something like Long Press on the raw.

the logic behind this is as we are going to rework things fundamentally soon (sub accounts typically), we need this to gets in even if not "prod ready".

@juan-cortes can you do the lil modif to have the access by long tap and then we send to QA?

juan-cortes:
@gre done, I found some weird operation list empty rows on tokens, might be worth checking
@Arnaud97234 can access the account distribution by long press on an asset from the asser distribution

PR #1052, opened 2019-10-02
llm now filter by marketcap

IAmMorrow:
add account's asset list now filter token + asset by marketcap. 

### Type

Feature

### Context

LL-1883

AddAccounts -> SelectCurrency -> CurrencyList

### DIFF
--- a/src/screens/AddAccounts/01-SelectCrypto.js
+++ b/src/screens/AddAccounts/01-SelectCrypto.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { Component } from "react";
+import React, { useMemo } from "react";
 import { translate, Trans } from "react-i18next";
 import { StyleSheet, View } from "react-native";
 // $FlowFixMe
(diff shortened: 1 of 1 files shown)


src/screens/AddAccounts/01-SelectCrypto.js
```diff
@@ -50,10 +53,20 @@ class AddAccountsSelectCrypto extends Component<Props, State> {
...
+    sortedCryptoCurrencies: []
+  }
+
   cryptocurrencies = listCryptoCurrencies(this.props.devMode).concat(
     listTokens(),
   );
 
+  async componentDidMount() {
+    const sortedCryptoCurrencies = await currenciesByMarketcap(this.cryptocurrencies)
+    this.setState({ sortedCryptoCurrencies })
```
gre: I kinda would prefer we use this hook: https://github.com/LedgerHQ/ledger-live-common/blob/master/src/currencies/sortByMarketcap.js#L52

because it already starts with the sorted list so it prevent a double rendering reordering flash.

gre:
@IAmMorrow

PR #1053, opened 2019-10-04
Bump to 1.1.0 to fix ios build

juan-cortes:
This allows to build on xcode preventing the error we were getting.

Should be tested on Android/iOS with a nano x to ensure the communication is still working over ble.

image


MortalKastor:
Can confirm this fixes ios building on my machine too 🙏

PR #1054, opened 2019-10-04
LL-1900 [iOS] fix enabling biometric auth

MortalKastor:
In the process of switching to our own fork of [react-native-touch-id](https://github.com/LedgerHQ/react-native-touch-id), we missed binary linking back the native lib in XCode project.
Curiously it seems it only broke *enabling* the biometric auth, users with the option enabled in a version prior to the regression were still able to unlock as usual.

### Type

Regression fix.

### Context

#1036

Enabling TouchID or FaceID on an iPhone should now work.

#1053 is needed for being able to build for testing

### DIFF
--- a/ios/ledgerlivemobile.xcodeproj/project.pbxproj
+++ b/ios/ledgerlivemobile.xcodeproj/project.pbxproj
@@ -804,7 +804,6 @@
 				00C302E91ABCBA2D00DB3ED1 /* libRCTNetwork.a in Frameworks */,
 				139105C61AF99C1200B5F7CC /* libRCTSettings.a in Frameworks */,
 				832341BD1AAA6AB300B99B32 /* libRCTText.a in Frameworks */,
-				8F3AB69021F8C163005F3E8C /* Lottie.framework in Frameworks */,
 				139FDEF61B0652A700C62182 /* libRCTWebSocket.a in Frameworks */,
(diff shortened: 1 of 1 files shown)


MortalKastor:
> _Bug found_
> 
> 1. After Enabling biometric auth, the splash screen is shown but the confirmation popup only appear after an action from the user (pressing home button for instance).
>    Without any action it can stays on splash screen for a while (like 10 sec).

I was unable to reproduce this behavior, in `dev` or `release` (staging) builds, on iPhone 6s Plus under iOS 12.4 or iPhone X under iOS 13.0

> 2. After Enabling biometric auth, when the confirmation modal is shown, if user use the wrong finger the following error is displayed: "Authentication fails..." instead of retry.

I was worried this happened when trying to _unlock_ when launching the app after faceID/touchID has been successfully enabled, and would leave the user stuck with no way to unlock with password, but that's not the case.

This issue you describe is clearly a suboptimal behavior, but is most probably not a regression from the current PR.

Also I just added a change which fixed a compilation issue (cyclic dependency) on my machine, can you check if this helps for your building issues @Arnaud97234 ?

gre:
thanks for investigation. we'll wait Arnaud to reproduce. In meantime we can merge that PR that fixes the linking issue.

PR #1059, opened 2019-10-11
[iOS] Move native dependencies to CocoaPods

MortalKastor:
The building of the iOS version of the app is quite brittle right now, and chokes on dependencies on some machine while working on others without a clear cause.

This is an attempt to fix that by moving the handling of this fragile part to [CocoaPods](https://cocoapods.org), which will be needed anyway for upgrading React Native.

### Type

Code Quality Improvement / Potential bug fix / Chore

### Parts of the app affected

The whole app on iOS.
This has the potential to break any feature relying on a lib that would have not been properly linked anymore.

### Caveat

CocoaPods needs to be installed on your machine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ yarn-error.log*
 third-party/
 
 xliffs/
+
+# Cocoa pods
+ios/Pods
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -1,0 +1,89 @@
+platform :ios, '9.0'
+
+target 'ledgerlivemobile' do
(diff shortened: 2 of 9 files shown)


gre:
let's merge this *as soon as* we are good with the 1.6.1 so it's not forgot for next release

MortalKastor:
These changes have been integrated, polished and extended in #1063, and since, in the end, switching to Cocoapods doesn't bring anything for us until we move to RN 0.61, this PR has no mean anymore.

gre:
sounds good.

PR #1062, opened 2019-10-22
Change from react-native-touch-id to react-native-fingerprint-scanner

juan-cortes:
Due to the old library seemingly being no longer maintained, this switch might be beneficial for us. Note that we are using an older version 2.6 than the latest since in order to benefit from the 3.0.x version we'd need to upgrade to react-native 0.60+ I believe @MortalKastor is working on this so once that part is done we could bump this library too.

This might be beneficial for the issues experienced by some users when using biometric authentication. 

I'm still figuring out some kinks (coming back to the app on ios does a weird placement of the prompt behind the logo screen for instance)

### Type

Code Quality Improvement

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -206,6 +206,8 @@ PODS:
     - React
   - react-native-config (0.11.7):
     - React
+  - react-native-fingerprint-scanner (3.0.2):
+    - React
   - react-native-locale (0.0.19):
     - React
   - react-native-netinfo (4.4.0):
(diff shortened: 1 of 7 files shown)


gre:
the PR is merged. what's the status now?

juan-cortes:
Rebased this branch with the latest develop and it seems like build take a fraction of what they used to so kudos to @MortalKastor @valpinkman and anyone involved really. I guess this still needs to go through QA.

This required passing from the 2.6.x branch to the 3.x branch as per https://www.npmjs.com/package/react-native-fingerprint-scanner#compatibility their compatibility guidelines.

I seem to have stumbled upon the reason for the bug that @Arnaud97234 was referring to a few weeks ago where the modal for the fingerprint scanner takes some time to show or never shows until you fail (use wrong finger). I initially thought it could be related to the lifecycle changes on io13 but then simply changing the event that triggered the showing of the privacy overlay seems to do the trick so 🤷‍♂

MortalKastor:
Small oversight, you forgot to remove `@ledgerhq/react-native-touch-id` otherwise lgtm 👍

gre:
so i guess just need to remove the forked version of the lib and this can be tested by Arnaud?

PR #1058, opened 2019-10-10
Bridge refactoring integration

juan-cortes:
### Type

Code Improvement

### Context

https://ledgerhq.atlassian.net/browse/LL-1922

### Parts of the app affected / Test plan

Send flow in general

PR #1065, opened 2019-10-30
Use the deleteLibcore function when resetting libcore.

phaazon:
This call will make future withLibcore calls to actually recreate the
Core structure.


gre:
i fixed the bug mentioned. will be part of #1058 
it's not perfect and hangs when you have no account but we'll improve later

PR #1071, opened 2019-11-07
Add the 1.6 product id to usb_device_filter.xml

juan-cortes:
### What is this
This pr is supposed to fix a bug where a nano s on version 1.6 is no longer detected when plugged into an android device. This is caused by us not having the correct product id listed in the usb_device_filter.xml file. However, there is a chance that this on its own is not enough and we might need to also include the product id at the @ledgerhq/ledgerjs level.

I'm unable to test this currently but if someone can build and make sure a 1.6 nanos is detected again on Android, then it works.


gre:
We'll get this in tomorrow

PR #1091, opened 2019-11-28
Fix keyboard covering bottom modal on iOS on select validator screen

MortalKastor:
It still doesn't make any sense to me to have a text input in a drawer, but at least it's working.

image

Android was handling this without issue, out of the box.

### Type

Bug fix

### Context

Tezos integrations

Adding a custom validator in delegation flow

### DIFF
--- a/src/components/InfoModal.js
+++ b/src/components/InfoModal.js
@@ -46,13 +46,14 @@ class InfoModal extends PureComponent<Props> {
       children,
       confirmLabel,
       confirmProps,
+      style,
     } = this.props;
     return (
       <BottomModal
         id={id}
         isOpened={isOpened}
         onClose={onClose}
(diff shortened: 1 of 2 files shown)


src/components/InfoModal.js
```diff
@@ -46,13 +46,14 @@ class InfoModal extends PureComponent<Props> {
...
       confirmProps,
+      style,
     } = this.props;
     return (
       <BottomModal
         id={id}
         isOpened={isOpened}
         onClose={onClose}
-        style={styles.modal}
+        style={[styles.modal, style || {}]}
```
gre: (minor: RN is totally fine with a falsy value in style)

MortalKastor: TIL 😯

PR #1106, opened 2019-12-03
Refactor biometrics authentication flow

MortalKastor:
Following #1062 biometrics auth was broken on Android, because while the former lib was shipping with a screen prompt for asking the user to touch the fingerprint reader on Android, the new one does not. iOS was not affected since this part is handled by the OS itself.

In order to fix that I had to create some UI for Android, which then required to refactor the logic.

The changes are more invasive than first expected, sorry 😞 

android screencap 2019-12-03 16:58:20

### Type

Regression fix.
Refactor.

### Context

#1062

- Enabling the the Fingerprint/Face ID option in Setttings screen
- Unlocking the app when the option is activated


gre:
seen with Arnaud, we assume this is acceptable as it's back to previous behavior. we can look more tomorrow.

PR #1107, opened 2019-12-03
Terms of Use

gre:
### Type

legal feature

### Context

legal

- trade safely is removed.
- term of use banner at first launch. (in future we can trigger again if terms change)
- the current content is not final, but this is dynamically fetch from github so we can replace with the content this week.
- blocking the user until accepted.
- resiliency: if no network (or bad api issue) it fallback on a link.


gre:
i think i have fixed that in the second commit

gre:
we'll merge so the build can be performed on CI

PR #1109, opened 2019-12-04
[iOS] Fix invisible chars in password field when OS is in dark mode

MortalKastor:
### Type

UI Polish

Password lock screen on iOS 13, with dark mode on


gre:
there might be other places where this problem happen. someone mentioned a problem on the placeholder color of amount field for instance.

MortalKastor:
> there might be other places where this problem happen. someone mentioned a problem on the placeholder color of amount field for instance.

I just added an app wide property to use Light theme colors, so that colors don't change when switching to Dark Mode. This should resolve all the issues related to native controls with no explicitly defined color.

PR #1111, opened 2019-12-07
Add gradientUnits to lineaGradient to fix regression

juan-cortes:
There was a regression in the graph chart that broke the lineargradient introduced when updating the `react-native-svg` library due to a less lenient rendering of the gradients that meant we'd show a solid color instead of the gradient. This fixes that

image

### Type

Regression

### Context

Slack

Graphs

### DIFF
--- a/src/components/Graph/DefGrad.js
+++ b/src/components/Graph/DefGrad.js
@@ -12,7 +12,14 @@ export default class DefGraph extends PureComponent<Props> {
   render() {
     const { height, color } = this.props;
     return (
-      <LinearGradient id="grad" x1="0" y1="0" x2="0" y2={height}>
+      <LinearGradient
+        id="grad"
+        x1="0"
+        y1="0"
+        x2="0"
+        y2={height}
+        gradientUnits="userSpaceOnUse"
+      >
         <Stop offset="0" stopColor={color} stopOpacity="0.3" />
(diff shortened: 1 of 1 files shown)


src/components/Graph/DefGrad.js
```diff
@@ -12,7 +12,14 @@ export default class DefGraph extends PureComponent<Props> {
   render() {
     const { height, color } = this.props;
     return (
-      <LinearGradient id="grad" x1="0" y1="0" x2="0" y2={height}>
+      <LinearGradient
+        id="grad"
```
gre: Are id globals?

juan-cortes: I could be wrong but I think the defs are global, hence the fix with unique ids we had to do elsewhere. However since nowhere in the app do we render more than one graph in the same screen it should be fine (?) 

This DefGrad essentially is a def used in the graph, worst case scenario it changes the gradient a bit. We can add a unique identifier for peace of mind later on. https://github.com/LedgerHQ/ledger-live-mobile/blob/d374d00b84465c040c9ce6a071b1ebd3129a071a/src/components/Graph/index.js#L82-L88

gre: Portfolio tab and account tab can be simultaneously rendered to me (as tabs content don't get unmounted)

PR #1112, opened 2019-12-09
Quickactions

juan-cortes:
This is an exploration of the quick-actions feature for both ios and android. The tldr for my end is that I'm not sure it's something worth polishing so it's best to put it out in the open for other eyes to see.

### What is this feature
This _quick actions_ are a context menu that pops when you long press - force touch on an app's icon, and it lists a bunch of items that should allow the user to fulfil an action faster than simply opening the app and navigating manually.

### What's the current status
It works, or at least it should work. It should even have a conditional item rendering for when in read only mode on iOS but there are some issues that prevent me from feeling confident in the feature.

### Current issues & Identified problems
- The launch doesn't seem to be any faster, although I don't know why I expected it to be faster.
- Linking to the tabs seems to provide a marginal gain if any, it's not like on an email app where I could save a bunch of clicks by having a _compose new email_ action. Here's it's basically the same number of touch events. The one that triggers the context menu + the selection versus the normal app entry + the tab selection.
- I've been unable to achieve a correctly displayed and sharp look for the icons, after multiple iterations with @bscholving and @dekkaki the icons still display a bit weirdly. In addition to this, Android has quicka-ction shortcuts where a user can potentially drag an action to the desktop and access it directly.
- When the app is locked the auto navigation will not navigate even after unlocking, this could probably be solved by handing out the navigation props over after the unlock.

### What could be explored to make the feature more appealing
- Linking to a particular account, I see a benefit in a power user that wants to check his favourite accounts without searching and going over all the accounts. But in order to get this we'd need to have searching and starred accounts on mobile first.
- Recurrent transaction or transaction templates, for instance a quick action that prefills the source account, amount and destination. But this doesn't feel safe because we need to verify the data anyway and what if the device is compromised...
- When we have privacy mode on mobile, we could have an option to open the app in privacy mode directly in order to skip the portfolio screen (to click the privacy mode) if someone else is with us. But once again, that feature is not implemented yet.

### Screenshots
image
Screen Shot 2019-12-09 at 13 13 46

### Type

UI Feature


juan-cortes:
Closing this after it's been idle for 58 days.
We will revisit in the future maybe.

PR #1117, opened 2019-12-12
Activate Hermes and upgrade Babel

juan-cortes:
This pr enables hermes engine for Android (depends on https://github.com/LedgerHQ/ledger-live-common/pull/420 to not break the displayed currency values) plus bumps a bit of everything. Leaving it as draft until @gre tells me what's going on with this.


gre:
Let's fully test and review this PR but not merge it yet (as we plan to release a new LLM before, but also we should release new version of ledgerjs/live-common when they get stable and confirmed it works)

cc @Arnaud97234

PR #1121, opened 2019-12-16
LL-2081 Flipper integration 🐬

MortalKastor:
Integrating [Flipper](https://fbflipper.com/), to get some more debug informations.

image

Not as useful as I hoped for the time being, but at least we get the full network requests and responses from libcore on iOS 🎉

### Usage

- Install [Flipper](https://fbflipper.com/) on your computer
- Launch it 🚀
- Run Ledger Live Mobile in debug as usual

### Caveats

- I found nothing interesting we didn't already have with out current tools for Android, but Flipper for RN on Android lags a bit behind its iOS counterpart, so it might get better with time
- It's currently only compatible with the simulator on iOS

### Type

Debug tool integration

### Context

https://ledgerhq.atlassian.net/browse/LL-2081

### Parts of the app affected

App building, so native dependencies as well

### Test plan

Since I had to trick a bit some XRP lib dependencies in order to prevent them to bundle an old lib over the one Flipper is shipping (cf scripts/post.sh), we need to make sure that didn't break XRP (especially on iOS production build)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -231,6 +231,15 @@ dependencies {
     } else {
         implementation jscFlavor
     }
+
+    implementation 'com.facebook.soloader:soloader:0.6.0+'
+
+    debugImplementation("com.facebook.flipper:flipper:${FLIPPER_VERSION}") {
+        exclude group:'com.facebook.yoga'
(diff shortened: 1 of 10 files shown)


gre:
@MortalKastor that looks super interesting. oculd you mention it in README.md?

PR #1131, opened 2020-01-15
LL-2093 Add missing description

juan-cortes:
Missing description 

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-2093

Partners screen

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -362,7 +362,7 @@
     "paybis": "Paybis.com is one of the most trusted cryptocurrency brokers worldwide since 2014, customer satisfaction - 4.7/5 via Trustpilot. Visa/Master cards, Bank Transfers, Skrill/Neteller accepted. 24/7 online customer support.",
     "shapeshift": "ShapeShift is an online marketplace where users can buy and sell digital assets. It is a fast and secure way for the world to buy and sell digital assets, with no lengthy signup process, no counterparty risk, and no friction.",
     "simplex": "Simplex is a EU licensed financial institution, providing a fraudless credit card payment solution.",
-    "swft": "SWFT Blockchain is a worldwide cross-blockchain transfer protocol and payment network, enabling simpler, faster and safer swaps between over 200 cryptocurrencies and free crypto payments to anyone, anywhere",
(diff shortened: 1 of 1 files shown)


src/locales/en/common.json
```diff
@@ -362,6 +362,7 @@
     "paybis": "Paybis.com is one of the most trusted cryptocurrency brokers worldwide since 2014, customer satisfaction - 4.7/5 via Trustpilot. Visa/Master cards, Bank Transfers, Skrill/Neteller accepted. 24/7 online customer support.",
     "shapeshift": "ShapeShift is an online marketplace where users can buy and sell digital assets. It is a fast and secure way for the world to buy and sell digital assets, with no lengthy signup process, no counterparty risk, and no friction.",
     "simplex": "Simplex is a EU licensed financial institution, providing a fraudless credit card payment solution.",
+    "swft": "SWFT Blockchain is a worldwide cross-blockchain transfer protocol and payment network, enabling simpler, faster and safer swaps between over 200 cryptocurrencies and free crypto payments to anyone anywhere",
```
guycaprio: There is a missing dot at the end of the sentence.

guycaprio: Other descriptions are in the same case: **luno**, **changenow**, **btcdirect** and **crypto**; can you please take the opportunity to update them also ?

guycaprio: Capture d’écran 2020-01-16 à 15 01 57
Otherwise UI test is OK.

juan-cortes: For these tasks I simply copy and paste what's on Jira so yeah @dasilvarosa you have the final word on this. I don't mind changing the rest to add a dot at the end either. Also, should we change them on desktop too?

dasilvarosa: It's indeed better with a full stop since these are full sentences. 👍

juan-cortes: @guycaprio the missing final dots are addressed in the already merged referenced issue, I'll rebase so that it's updated here once @dasilvarosa lets us know if we want "anyone anywhere" or "anyone, anywhere" 😪

juan-cortes:
So with the rebased develop and the confirmation of the latest wording, I think we are good to go finally. @gre

guycaprio:
A final dot and it will be merged :trollface:

PR #1133, opened 2020-01-15
Remove bitpanda from partners

juan-cortes:
Let's see how small of a PR I can end up doing :trollface: 

### Type

Copy

### Context

https://ledgerhq.atlassian.net/browse/LL-2108


gre:
damn bot

juan-cortes:
Yeah, we need a `live-common` bump first to make it official.

gre:
Please provide a PR that bumps both latest live-common and libcore. We need to assess all is good first and they both need to be together. Thx

juan-cortes:
Ok thx :trollface:

juan-cortes:
Rebasing latest develop brought the addition of swft and it had no description either.
So, linking this to lld-v2 LL-2093 I've added the same description. Note that I have not added the extra "," as suggested by @guycaprio but I agree that perhaps that should be added. Still waiting for @dasilvarosa  on those issues

juan-cortes:
This and https://github.com/LedgerHQ/ledger-live-mobile/pull/1131 need to be both merged. The periods referred to https://github.com/LedgerHQ/ledger-live-mobile/pull/1131#discussion_r368918531 are addressed in this one and they've been pending for way too long imo @gre

PR #1138, opened 2020-01-17
Common + core bumps

juan-cortes:
After bumping live-common and the rest of the ledgerhq libraries I noticed that the asset distribution screen was crashing for some reason. While I fixed (more or less) it by removing the globalSync wrapper for that list, I think the app feels janky and sluggish but I don't know if it's something that has been happening for a while or not.

@Arnaud97234 @gre


gre:
@juan-cortes can you upgrade again the libs? there is a fix for Tezos scanning issue.

PR #1141, opened 2020-01-27
update live-common@11.3.0

gre:
This should fixes the Tezos issue with KT disappearing @Arnaud97234 

also includes latest ERC20 and bakers list.


gre:
@Arnaud97234 fixed

PR #1143, opened 2020-01-29
Bump live-common with latest socket implementation

gre:
For the test plan, Please refer to
https://github.com/LedgerHQ/lld-v2/pull/164

it will be similar but we need to test in both iOS and Android because known implementation difference we noticed in the past.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "@ledgerhq/hw-app-xrp": "5.7.0",
     "@ledgerhq/hw-transport": "5.7.0",
     "@ledgerhq/hw-transport-http": "5.7.0",
-    "@ledgerhq/live-common": "^11.3.0",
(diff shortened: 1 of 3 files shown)


guycaprio:
Nano X (latest firmware) on Android & iOS:

- [x] Refusing allow manager
- [x] Disconnecting device in various time of the flow (lockscreen + bluetooth off)
- [x] Opening the app in the middle of the flow
- [x] Interrupting the entering of manager and coming back to it
- [x] interrupting the apps installation
- [x] network down at the beginning of the flow
- [x] network down in the middle => Connection failed (websocket error)
---
ccaf77b2 (HEAD -> live-common-socket-rework, gre/live-common-socket-rework) Bump live-common with latest socket implementation

PR #1144, opened 2020-01-30
LL-2098: visual issue on small devices with keyboard open on amount input

LFBarreto:
An visual issue occurred on small devices when trying to send currencies on step 3. while keyboard was opened the amount input section was off and cropped

the accounts input can now shrink up to 50% of initial height and has a normalized font size calculated according to screen size.

### Type

Bug Fix

### Context

LL-2098

Amount input on send flow step 3


src/logic/normalizeSize.js
```diff
@@ -0,0 +1,14 @@
+import { Dimensions, Platform, PixelRatio } from "react-native";
+
+const { width: SCREEN_WIDTH } = Dimensions.get("window");
```
gre: can you use 

```
import getWindowDimensions from "./getWindowDimensions"
```

?

src/screens/SendFunds/AmountInput.js
```diff
@@ -27,6 +27,8 @@ import CurrencyInput from "../../components/CurrencyInput";
 import TranslatedError from "../../components/TranslatedError";
 import type { T } from "../../types/common";
 
+import normalizeZize from "../../logic/normalizeSize";
```
gre: minor typo

src/screens/SendFunds/AmountInput.js
```diff
@@ -166,24 +168,25 @@ const styles = StyleSheet.create({
...
   wrapper: {
-    minHeight: 100,
+    flexBasis: 100,
+    flexShrink: 0.5,
     flexDirection: "column",
     justifyContent: "center",
   },
   currency: {
-    fontSize: 24,
+    fontSize: normalizeZize(20),
```
gre: I'm not sure of variable font size idea. we don't do that anywhere else and that would feel weird if part of the app start doing that. I would like to hear from design too.

to me, fixing by flex should be enough, it's not?

LFBarreto: Just fixed the issue without the need to normalize text font size. I figured that this particular active font size was rather big but it can fit in the 5S screen just right if the other elements don't grow out of flex space :+1:

PR #1145, opened 2020-02-03
LL-2066 SubAccounts header on operation details

juan-cortes:
image

Since the information circle triggered a modal that was aimed at erc20 tokens specifically I removed it until we have a better solution or different link/content for subaccounts. Ping @aramab 

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-2066

On the operation details page of an operation containing subaccount operations (think Tezos) the section that is shown above as "Subaccount operations" would read "Token operations".

### DIFF
--- a/src/screens/OperationDetails/Content.js
+++ b/src/screens/OperationDetails/Content.js
@@ -18,6 +18,7 @@ import uniq from "lodash/uniq";
 import { connect } from "react-redux";
 import { createStructuredSelector } from "reselect";
 import { Trans } from "react-i18next";
+import { listTokenTypesForCryptoCurrency } from "@ledgerhq/live-common/lib/data/tokens";
 import { localeIds } from "../../languages";
 import LText from "../../components/LText";
 import OperationIcon from "../../components/OperationIcon";
@@ -88,6 +89,8 @@ class Content extends PureComponent<Props, State> {
   render() {
     const { account, parentAccount, operation, currencySettings } = this.props;
     const mainAccount = getMainAccount(account, parentAccount);
+    const isToken =
+      listTokenTypesForCryptoCurrency(mainAccount.currency).length > 0;
     const unit = getAccountUnit(account);
     const parentUnit = getAccountUnit(mainAccount);
     const currency = getAccountCurrency(account);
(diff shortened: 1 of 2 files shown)


src/screens/OperationDetails/Content.js
```diff
@@ -88,6 +89,8 @@ class Content extends PureComponent<Props, State> {
   render() {
     const { account, parentAccount, operation, currencySettings } = this.props;
     const mainAccount = getMainAccount(account, parentAccount);
+    const isToken =
+      listTokenTypesForCryptoCurrency(mainAccount.currency).length > 0;
```
MortalKastor: I'm getting confused... isn't what the UI call "subaccount" really means the `ChildAccount` type?

Wouldn't that break in case we integrate a currency supporting both tokens and subaccounts/childaccounts?

Isn't it possible to just check for `account.type === "TokenAccount"` (or we have some accounts typed as `TokenAccount` which should be displayed as _subaccounts_?)

juan-cortes: I mimicked the logic from lldv2 https://github.com/ledgerhq/lld-v2/blob/820fd4e3d4f1ecab1bbfd16a5d3216faac350e1c/src/renderer/modals/OperationDetails/index.js#L327 which ironically comes from you 🤣

MortalKastor: I'm not really the author of that line, I just did the v1 to v2 copypasta :trollface: 

@valpinkman I'm probably missing something, but I don't get why you did `listTokenTypesForCryptoCurrency(mainAccount.currency).length > 0` rather than `account.type === "TokenAccount"` to check if an account is a `Token` in
https://github.com/LedgerHQ/ledger-live-desktop/blob/master/src/components/modals/OperationDetails.js#L243

I this some kind of pattern we should continue to follow?
Shouldn't `account.type` always be trusted? 😕

juan-cortes: oh sorry for that, I trusted the git _blame_

MortalKastor: My bad, I thought the check was at the child level rather than on the parent,  don't mind me, I'm going burry myself.

PR #1150, opened 2020-02-10
 LL-1888 debounce clicks on navigations push

LFBarreto:
as mentioned in #1146 
we should prevent double click navigation pushs

best solution yet is to debounce them.

### Type

Bug Fix

### Context

LL-1888

Accounts tokens links and sub operations aswell


juan-cortes:
cc @Arnaud97234 @nabil-brn

PR #1127, opened 2020-01-02
Ll 2075 manager v2

LFBarreto:
Manager V2

### Type

Feature

### Context

issue LL-2075

### Parts of the app affected / Test plan

Whole manager screens

PR #1151, opened 2020-02-12
Update to live-common 12

gre:
### What to expect

- That nothing changes! Everything should work the same.
- the netwoking logic is now in live-common to share more code.
- account#starred field. not yet used but it's ready to be used. cc @juan-cortes 
- apps size polyfill data has been dropped (as it's not exposed by the api).

### What to **not** expect

- "locked" mode of the device is not yet properly recovered (in context of manager), but it's in progress. it's a complex topic.


gre:
this will be part of #1127

PR #1153, opened 2020-02-17
Allow TypeScript files

JunichiSugiura:


gre:
we'll reopen when we first migrate live-common

PR #1161, opened 2020-02-26
bump live-common 12.0.x

LFBarreto:
bump live common version to 12.0.x version


gre:
we'll merge to make sure it goes into tonight build and directly test with the testflight build. If QA determine there is problem, we can always address tomorrow.

PR #1181, opened 2020-03-19
LL-2254 Info modal

LukeSugiura:


src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -1,118 +1,130 @@
...
+  }, []);
+
+  const onPressInfoCreator = useCallback(
+    (infoName: InfoName) => () => setInfoName(infoName),
+    [],
+  );
+
+  const isModalOpen = useMemo(() => !!infoName, [infoName]);
+
+  const data = useMemo(() => (infoName ? infoCandidates[infoName] : []), [
```
LFBarreto: these memo are a bit overkill in my opinion you could just set data to `const data = infoCandidates[infoName] || [];`
and useMemo for a direct boolean value on isModalOpen also is a bit much :octocat:

src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -1,118 +1,130 @@
...
-          <Info size={12} color={colors.grey} />
-        </View>
-        <LText semiBold style={styles.balance}>
-          {formatedEnergy || "–"}
-        </LText>
-      </TouchableOpacity>
+      <InfoModal isOpened={isModalOpen} onClose={onCloseModal} data={data} />
+
+      <InfoItem
+        title={t("account.availableBalance")}
```
LFBarreto: i don't think using and connecting the translation function here is necessary you could just use a Trans component thus limiting higher order functions usage.
like so `title={<Trans i18nKey="account.availableBalance" />}`
if theres an issue on typing instead of using string you can use React$Node for flow. 

Alternatively you can send the translation key instead of the translated string and use a Trans component inside InfoItem :man_shrugging:

src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -150,4 +162,41 @@ const styles = StyleSheet.create({
   },
 });
 
-export default AccountBalanceSummaryFooter;
+function getInfoCandidates(t: TFunction): { [key: InfoName]: ModalInfo[] } {
+  const currency: CryptoCurrency = (findCryptoCurrencyById("tron"): any);
```
LFBarreto: all of this could be put inside const variables outside of a function or component scope since its called in the component without a memo it will be reprocessed i guess :man_shrugging:

src/icons/Freeze.js
```diff
@@ -2,16 +2,23 @@
...
-      fill={color}
-    />
-  </Svg>
-);
-
-export default Freeze;
+export default function Freeze({
+  size = 16,
+  color = colors.black,
```
LFBarreto: i think we use the darkBlue color instead of black on all the "black" text instead :+1:

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: if you upgrade react-i18n, they have handy hook called "useTranslation". I tried to use that but apparently, it doesn't exist in this version so I went with HOC. It is just a temporary solution.

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: oh yea, I was using memo before but I changed this func to custom hook but useTranslation doesn't exported so I put it back and forgot to useMemo with this func.

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: yup, I agree. I still have to stick with ternary operator due to its type tho.

src/families/tron/AccountBalanceSummaryFooter.js
gre: `<Trans` should work regardless of react-i18n version and we generally prefer it

src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -1,118 +1,127 @@
...
+      />
+      <InfoItem
+        title={t("account.bandwidth")}
+        onPress={onPressInfoCreator("bandwidth")}
+        value={formattedBandwidth.toString() || "–"}
+      />
+      <InfoItem
+        title={t("account.energy")}
+        onPress={onPressInfoCreator("energy")}
+        value={formattedEnergy.toString() || "-"}
```
gre: I imagine this is consistent with desktop? I think this should be a formatter in live-common. who knows if we don't want to actually format it in better ways, like with commas? just the fact we have `–` kinda make me wonder maybe we want to factorize this 🤔

src/families/tron/AccountBalanceSummaryFooter.js
gre: please use `getCryptoCurrencyById` which is the "less safe" equivalent but that would prevent you to cast as `any` and ignore the induced flow error. agreed with the outside scope.

src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -150,4 +159,41 @@ const styles = StyleSheet.create({
   },
 });
 
-export default AccountBalanceSummaryFooter;
+function getInfoCandidates(t: TFunction): { [key: InfoName]: ModalInfo[] } {
+  const currency: CryptoCurrency = (findCryptoCurrencyById("tron"): any);
+  const TronIcon: React$ComponentType<{
+    color: string,
+    size: number,
+  }> = (getCryptoCurrencyIcon(currency): any);
```
gre: not fan of this pattern to ignore the type with a `any` cast. sounds like we are hiding flow errors and we shouldn't. I prefer // $FlowFixMe

src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -150,4 +159,41 @@ const styles = StyleSheet.create({
...
+        Icon: () => <BandwidthIcon size={18} />,
+        title: t("tron.info.bandwidth.title"),
+        description: t("tron.info.bandwidth.description"),
+      },
+    ],
+    energy: [
+      {
+        Icon: () => <EnergyIcon size={18} />,
+        title: t("tron.info.energy.title"),
+        description: t("tron.info.energy.description"),
```
gre: it sounds like these could all be statically defined if you were using `<Trans`

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: I prefer `const {t} = useTranslation();` in general because it's simply returns string (also the type).

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: I cast as any and also specified const type due to Flow's limitation and its workaround.

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: Cool! i didn't know about this function (getCryptoCurrencyById). I'll replace. I casted as any with the same reason I described here (https://github.com/LedgerHQ/ledger-live-mobile/pull/1181#discussion_r396332325).

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: https://react.i18next.com/latest/trans-component#important-note

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: we'll discuss this in other issue 👍

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: Stick with good old t func that simply returns string for now.

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: fixed: https://github.com/LedgerHQ/ledger-live-mobile/pull/1181/commits/a4ea5d67fa515395176b5d7c606385f95c5f105c

src/families/tron/AccountBalanceSummaryFooter.js
LukeSugiura: fixed: https://github.com/LedgerHQ/ledger-live-mobile/pull/1181

src/families/tron/AccountBalanceSummaryFooter.js
```diff
@@ -1,118 +1,127 @@
...
   alwaysShowSign: false,
   showCode: true,
 };
 
-const AccountBalanceSummaryFooter = ({ account }: Props) => {
-  if (!account.tronResources) return null;
+type InfoName = "available" | "frozen" | "bandwidth" | "energy";
+
+function AccountBalanceSummaryFooter({ account, t }: Props) {
+  const [infoName, setInfoName] = useState<InfoName | typeof undefined>();
```
gre: `InfoName | typeof undefined` => `?InfoName`

LukeSugiura: yea it can be, but personally I rather not using null. When we transform to TS in the future, there's no maybe type so we need to define as `InfoName | undefined | null`

PR #1206, opened 2020-04-02
[iOS] Fix building for iDevices w/o breaking simulator

MortalKastor:
- Removes everything related to the libcore `x86` _thin_ framework
- Uses the new libcore _fat_ framework featuring all the archs, including `x86`
- Adds a script to strip from the _fat_ framework any arch not matching the current target (in the embedded copy, not the project folder one)
- Adds a new `Staging` configuration which is a copy of `Release` with `ONLY_ACTIVE_ARCH` set to `YES` so that it doesn't try to build for `i386` when running in the simulator

HODL for now because we need both LedgerHQ/lib-ledger-core#545 and LedgerHQ/lib-ledger-core-react-native-bindings#53 to be merged for it to work (it's using LedgerHQ/lib-ledger-core-react-native-bindings#52 as a POC right now).

### Type

Bug Fix

### Context

Surprise XCode update, breaking our build script.
LedgerHQ/lib-ledger-core#545
LedgerHQ/lib-ledger-core-react-native-bindings#53

- [x] Dev mode works on device
- [x] Staging works on device
- [x] Dev mode works on simulator
- [x] Staging works on simulator


valpinkman:
tested this morning with @MortalKastor 

- tested on dev on real device (works, reported by me)
- tested on staging on real device (works, reported by me)
- tested on dev on simulator (works, reported by thibaut)
- tested on staging on simulator (still breaks, reported by thibaut)

I think it's ready enough so we can merge and start building again

wdyt @gre @MortalKastor @Arnaud97234

MortalKastor:
The only thing is that it's using a custom build of `react-native-ledger-core` right now (`4.10.1-dev.0`). I need to change a bit how `libcore.framework` is built on libcore repo (rather than messing with files in the preinstall of the bindings like I did) to make a proper build that will work with the changes from the current PR (and probably fix staging on simulator along the way).

The PR doesn't really need to wait for the incoming libcore changes to be in though, as long as we're fine with targeting a `react-native-ledger-core` version that doesn't follow the state of it's repo in the meantime.

MortalKastor:
Updated comment to reflect current state

gre:
@Arnaud97234 can you test this again on a real device?

I merged develop and also bump latest libcore that allows this PR to get merged 🙏

PR #1207, opened 2020-04-03
LL-2207 Blacklisting tokens feature

juan-cortes:
### Description
We introduce the possibility to blacklist a token and hide all its related accounts, this was a much-requested feature since the appearance of junk tokens that users didn't want to see in their portfolios. This is the port of the recently merged (even though the text has not been reviewed by @dasilvarosa yet) https://github.com/LedgerHQ/ledger-live-desktop/pull/2788 with the peculiarities that come from it being mobile.

We can access this bottom drawer modal from three different places:
- In a parent account by long press on a token account row listed
- In an asset screen for a token, the top-right header action now doesn't directly show the contract information but rather an intermediary modal with the possibility of hiding the token or showing the contract.
- In the general account list by long-pressing the row for a token.

We can show these accounts again by visiting the new section inside `Settings` called `Accounts` where we moved also the option to hide empty token accounts. We are missing a description for that row too
image

### Screenshots

diagram

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2207
https://www.figma.com/file/KdPZlnD9BOpwvaor4twU0H/Hidden-Token?node-id=72%3A46

Make sure we can blacklist and show again tokens, just like on https://github.com/LedgerHQ/ledger-live-desktop/pull/2788
As a side note, probably something that can be improved, I was unable to trigger the sync automatically so showing again the tokens will require the user to manually pull down on the screen. I'm sure @gre or someone else can point me in the right direction to sort that one.


src/locales/en/common.json
```diff
@@ -860,6 +860,15 @@
       "currencySettingsTitle": "{{currencyName}} Settings",
       "placeholder": "No settings for this asset"
     },
+    "accounts": {
+      "header": "Accounts",
+      "title": "Accounts",
+      "desc": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut nisl lorem.",
```
gre, line 866: can't merge this

src/live-common-setup.js
```diff
@@ -12,8 +12,8 @@ import { registerTransportModule } from "@ledgerhq/live-common/lib/hw";
 import type { TransportModule } from "@ledgerhq/live-common/lib/hw";
 import BluetoothTransport from "./react-native-hw-transport-ble";
 import "@ledgerhq/live-common/lib/load/tokens/ethereum/erc20";
-// import "@ledgerhq/live-common/lib/load/tokens/tron/trc10";
-// import "@ledgerhq/live-common/lib/load/tokens/tron/trc20";
```
gre, line 16: shouldn't be this PR's concern to enable these. will be somewhere else.

src/reducers/settings.js
```diff
@@ -86,6 +87,7 @@ const INITIAL_STATE: SettingsState = {
   experimentalUSBEnabled: false,
   countervalueFirst: false,
   hideEmptyTokenAccounts: false,
+  blacklistedTokenIds: ["ethereum/erc20/0x_project"],
```
gre, line 90: why?

src/screens/Settings/Accounts/index.js
```diff
@@ -0,0 +1,172 @@
...
+  bottom: 16,
+};
+
+class AccountsSettings extends PureComponent<Props> {
+  static navigationOptions = {
+    title: i18next.t("settings.accounts.title"),
+  };
+
+  componentWillUnmount() {
+    // TODO trigger a refresh
```
gre, line 49: 🤔

src/locales/en/common.json
juan-cortes: Give me a better temporary wording and I'll update

src/reducers/settings.js
juan-cortes: Because I finished at 4 am and I wasn't thinking straight

src/screens/Settings/Accounts/index.js
juan-cortes: Will be addressed once the sync rework is completed on https://github.com/LedgerHQ/ledger-live-mobile/pull/1208 and will look exactly like on desktop

src/live-common-setup.js
juan-cortes: Fair enough

src/screens/Settings/Accounts/BlacklistTokenModal.js
```diff
@@ -0,0 +1,152 @@
...
+import { Trans } from "react-i18next";
+import Icon from "react-native-vector-icons/dist/Feather";
+import BottomModal from "../../../components/BottomModal";
+import LText from "../../../components/LText";
+import CurrencyIcon from "../../../components/CurrencyIcon";
+import BanIcon from "../../../icons/Ban";
+import Touchable from "../../../components/Touchable";
+import { blacklistToken } from "../../../actions/settings";
+import colors from "../../../colors";
+import AlertTriangle from "../../../icons/AlertTriangle";
```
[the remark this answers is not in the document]
juan-cortes: addressed, I missed it in the hundreds of warnings

juan-cortes:
> The only thing to rework is that the "x" button hitbox to remove a token from the blacklist is too big, when clicking on the token name it deletes it from the list.

The issue addressed in the current version @nabil-brn

PR #1208, opened 2020-04-03
Migrate bridge sync (perf) + useTronPowerLoading logic

gre:
- This should improve a lot performance of syncs as well as the one from 🌳 .
- it also bring the fixes for tron power to be ready after the timer
- it needs important QA check around adding accounts and synchronizing them, but the logic has been proven to work on desktop as it was merged 2 weeks ago there.


src/screens/FreezeFunds/04-ValidationSuccess.js
```diff
@@ -51,6 +34,7 @@ type Props = {
 
 const ValidationSuccess = ({ account, navigation }: Props) => {
   const time = useTimer(60);
```
LFBarreto, line 36: if we're not using this we should remove it :+1: since you remove the useTimer as well

gre: we are using it 🤔

PR #1209, opened 2020-04-03
Minimize re-render, Truncate long SR names, add Selected section

LukeSugiura:


LFBarreto:
merging this i'll check what we can do in order to make it less laggy on the inputs :+1:

PR #1210, opened 2020-04-03
Fallback to USD when the user selected fiat counterValue is not supported

juan-cortes:
### Description
This pr introduces the automatic fallback to using USD as a counter value when the currently selected counter value is not reported as supported. Without this, users who have selected a non-supported fiat counter value will see their total balances and graphs as broken.

### Type
Feature

### Context
Slack conversation, no Jira task as far as I know

Parts of the app affected / Test plan
- With the **desktop** application stopped and before this https://github.com/LedgerHQ/ledger-live-desktop/pull/2795 gets merged
- Edit your `app.json` and set the counter value to something not supported (but contemplated in https://github.com/LedgerHQ/ledger-live-common/blob/1a29a3bc5f533ee29e0a47c577a1a7e068546e0b/src/data/fiat.js#L23-L73. 
- Launch the app, open the console, and export the accounts/settings. You'll see a `BRIDGESTREAM_DATA` variable in the console.
- Copy that into your mobile's project `.env` file. 
- Launch the mobile app, go into the `Settings>Debug>Mock>Import` and import the settings. It should set the counter value to the invalid one from the desktop.
- Move to this branch if not here already, and relaunch the app. It should fix the broken portfolio graph and set the counter value to USD again.

For best results, use an `app.json` with a single account since large exports sometimes don't work with the emulator when it's an env.

### DIFF
--- a/src/context/LedgerStore.js
+++ b/src/context/LedgerStore.js
@@ -10,6 +10,7 @@ import reducers from "../reducers";
 import { importSettings } from "../actions/settings";
 import { importStore as importAccounts } from "../actions/accounts";
 import { importBle } from "../actions/ble";
+import { INITIAL_STATE, supportedCountervalues } from "../reducers/settings";
 
 const createLedgerStore = () =>
   createStore(
@@ -55,6 +56,15 @@ export default class LedgerStoreProvider extends Component<
     store.dispatch(importBle(bleData));
 
     const settingsData = await db.get("settings");
+    if (
+      settingsData &&
+      settingsData.counterValue &&
+      !supportedCountervalues.find(
+        ({ ticker }) => ticker === settingsData.counterValue,
+      )
+    ) {
+      settingsData.counterValue = INITIAL_STATE.counterValue;
+    }
     store.dispatch(importSettings(settingsData));
 
     const accountsData = await db.get("accounts");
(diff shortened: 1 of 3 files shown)


juan-cortes:
Basically we can't rely only on fiat supported currencies since there are some other cryptocurrency that can be used. I'm putting this on hold and will look at it on monday.

juan-cortes:
Changed the logic so that it fallback to USD or whatever is set in `initialSettings` but also allows the selection of `supportedCryptoCurrencies` as countervalues, at this point, it should allow BTC and ETH if I'm not mistaken. Testing should be the same as described above.

gre:
To test it, in the app, select a fiat, like CAD,

edit the file `node_modules/@ledgerhq/live-common/lib/data/fiat.js` and remove that fiat in the array of

```
setSupportedFiats(["CAD", "CHF", "EUR", "GBP", "HKD", "ILS", "JPY", "KRW", "MXN", "PLN", "RUB", "SGD", "TRY", "USD"]);
```

you can turn it into 


```
setSupportedFiats(["CHF", "EUR", "GBP", "HKD", "ILS", "JPY", "KRW", "MXN", "PLN", "RUB", "SGD", "TRY", "USD"]);
```

and then **deploy again**

@nabil-brn

PR #1211, opened 2020-04-06
Use high resolution image for Earn rewards screen

LukeSugiura:


src/families/tron/IlluRewards.js
```diff
@@ -3,8 +3,8 @@ import React from "react";
 import { Image } from "react-native";
 import illuRewardsPng from "../../images/rewards.png";
 
-const IlluRewards = ({ style }: *) => (
-  <Image style={style} source={illuRewardsPng} width={150} height={150} />
+const IlluRewards = ({ style }: any) => (
+  <Image style={style} source={illuRewardsPng} width={230} height={230} />
```
LFBarreto, line 7: have you tried this on a small height device i was worried about having a scroll on this view if we are on a 5S :thinking:

PR #1212, opened 2020-04-06
use estimateMaxSpendable on mobile

valpinkman:
Send Flow => Estimate is now correct instead of using account.balance, we use estimateMaxSpendable

Tested on iOS, seems to work


gre:
lgtm to merge to the main PR @LFBarreto

PR #1222, opened 2020-04-07
Prevent goBack to Started page if it is already skipped

JunichiSugiura:
If you try managing votes from an account that already voted
and then navigate back using native navigation
you should be redirected to the account page now instead of the Starter info one.


gre:
@LukeSugiura please provide more info on how to test this.

JunichiSugiura:
This bug is not longer relevant since current casting flow doesn't show back button anymore.

PR #1225, opened 2020-04-07
(TRX): contract address modal don't show if no url available

LFBarreto:
Updated contract address modal to not show if no url available.

### Type

UI Polish

### Context

Trx trc10 tokens

Tokens contract info modal

### DIFF
--- a/src/screens/Account/AccountHeaderRight.js
+++ b/src/screens/Account/AccountHeaderRight.js
@@ -6,6 +6,12 @@ import { connect } from "react-redux";
 import Icon from "react-native-vector-icons/dist/FontAwesome";
 import type { Account, TokenAccount } from "@ledgerhq/live-common/lib/types";
 
+import {
+  getDefaultExplorerView,
+  getAccountContractExplorer,
+} from "@ledgerhq/live-common/lib/explorers";
+import { getMainAccount } from "@ledgerhq/live-common/lib/account/helpers";
+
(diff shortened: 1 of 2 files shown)


src/screens/Account/AccountHeaderRight.js
```diff
@@ -39,6 +45,15 @@ const AccountHeaderRight = ({ navigation, account, parentAccount }: Props) => {
...
+
+    const url = getAccountContractExplorer(
+      explorerView,
+      account,
+      parentAccount,
+    );
+
     return (
       <>
         <Touchable event="ShowContractAddress" onPress={toggleModal}>
```
gre, line 59: mmh, i might be wrong, but shoudln't we return null completely if there is no url? that's what i meant by "not showing it at all if it's not possible"

LFBarreto: we have a modal that asks to hide the token as well we have to be able to hide the token except the modal showing the contract address link won't show :+1:

gre: ok

gre:
Also we are not supposed to conditionally render a modal to make it appear. It's better to use the isOpened Boolean for that. Am I seeing something wrong here? cc @juan-cortes

juan-cortes:
The first modal appearing or not was fixed on the other pr I did, as to conditionally rendering them or not, it seemed cleaner this way but I haven't done mobile in a while maybe you're right

PR #1226, opened 2020-04-07
Fix to reset modal confirmation for blacklisting tokens

juan-cortes:
This pr resets the state of the blacklist modal every time it's closed, before it will only show the first step the first time.


juan-cortes:
I've tested this now and it works.

PR #1228, opened 2020-04-08
(trx): token operation details after send success issue

LFBarreto:
operation details afetr a sent of token now redirects to the suboperation if available

Screenshot_2020-04-08-09-43-59-927_com ledger live debug

### Type

UI Polish

Send a transaction from a token to a token and click on the preview of the operation details to see the correct translation wording


gre:
essentially to me the fix should be:

if there is a subOperations[0], then prefer using the .id of it , otherwise use the parent’s
for opening the “view operation” thingy

PR #1236, opened 2020-04-09
(TRX): fixed issue on cast votes modal and keyboard open

LFBarreto:
Issue was occurring on cast votes modal when keyboard was open

should be fixed now:

Screenshot_1586426969

### Type

UI Polish

Trx account that can vote
try and edit a vote.


gre:
to me usage of KeyboardView is mandatory on iphone.

PR #1239, opened 2020-04-09
(TRX): fix send receive button on trc tokens having scroll

LFBarreto:
issue occuring on trc tokens send receive butons having scroll

### Type

UI Polish

Go to a trc token page you should see same send receive buttons as normal currencies.

### DIFF
--- a/src/screens/Account/AccountActions.js
+++ b/src/screens/Account/AccountActions.js
@@ -44,8 +44,6 @@ const AccountActions = ({
   const ReceiveAction =
     (decorators && decorators.ReceiveAction) || ReceiveActionDefault;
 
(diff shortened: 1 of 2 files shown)


src/screens/Account/AccountActions.js
```diff
@@ -56,6 +54,12 @@ const AccountActions = ({
     [accountId, parentId, navigation],
   );
 
+  const ManageAction =
```
gre, line 57: this shouldn't have a uppercase first character if it's not a component.

LFBarreto: fixed the manageAction component actually returns null in the case of trc tokens however a component returning null isn't falsy and i can't check against it in order to apply the right layout here :disappointed:

PR #1246, opened 2020-04-14
LL-2295 Android SQLite issue

LFBarreto:
(DB): updated accounts storage saving process to split the data by account key and not grouped

SQLite accounts data is now stored by account keys thus removing certain limitation of storage we can have on android

### Type

Bug Fix

### Context

LL-2295

Several tests needed on both Android and IOS:
- a cold start from the app with no previous data try import accounts and restart the app
- try start the app with previously set accounts from a different app version
- try remove accounts

more info on the main issue

### DIFF
--- a/src/db.js
+++ b/src/db.js
@@ -1,4 +1,116 @@
 // @flow
 import store from "react-native-simple-store";
 
-export default store;
+const ACCOUNTS_KEY = "accounts";
+const ACCOUNTS_DB_PREFIX = "accounts.active.";
+
+export default class DB {
+  static get(key: string | any[]): Promise<any> {
+    /** handle accounts case */
+    if (key === ACCOUNTS_KEY) return DB.getAccounts();
+    return store.get(key);
+  }
+
+  static save(key: string | any[], value: any): Promise<any> {
+    /** handle accounts case */
+    if (key === ACCOUNTS_KEY) return DB.saveAccounts(value);
+    return store.save(key, value);
+  }
+
+  static update(key: string, value: any): Promise<any> {
+    return store.update(key, value);
+  }
+
+  static keys(): Promise<any> {
+    return store.keys();
+  }
+
+  static delete(key: string | any[]): Promise<any> {
+    return store.delete(key);
+  }
+
+  static push(key: string, value: any): Promise<any> {
+    return store.push(key, value);
+  }
+
+  /** format account id to its account DB key */
+  static formatAccountDBKey = (id: string): string =>
+    `${ACCOUNTS_DB_PREFIX}${id}`;
+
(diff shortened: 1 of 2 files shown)


src/db.js
```diff
@@ -1,4 +1,87 @@
...
+      /** filter through them to get only the accounts ones */
+      const accountKeys = keys.filter(
+        key => key.indexOf(ACCOUNTS_DB_PREFIX) === 0,
+      );
+      /** if present return them */
+      if (accountKeys && accountKeys.length > 0)
+        return store.get(accountKeys).then(active => ({ active }));
+
+      /** else return old accounts data */
+      return store.get(ACCOUNTS_KEY);
```
gre, line 84: is this resilient to potentially account key that is getting dropped? If not, we need to protect atomicity of that logic VS set accounts.

src/db.js
```diff
@@ -1,4 +1,87 @@
...
+  }
+
+  static push(key: string, value: any): Promise<any> {
+    return store.push(key, value);
+  }
+
+  static getAccountDbKey = (id: string): string => `${ACCOUNTS_DB_PREFIX}${id}`;
+
+  /** save accounts method between SQLite db and redux store persist */
+  static saveAccounts({
```
gre, line 39: this needs to happen in an atomic way. there are many ways it could go wrong if multiple saveAccounts happen simultaneously.

src/db.js
```diff
@@ -1,4 +1,87 @@
...
+    active: newAccounts,
+  }: {
+    active: any[],
+  }): Promise<any> {
+    return DB.getAccounts().then(oldAccounts => {
```
gre, line 44: we shouldn't DB.getAccounts() we only need to get the keys for better perfs as proven by fact you do `.id`

fetching all is expensive it seems?

src/index.js
```diff
@@ -61,7 +62,12 @@ class App extends Component<*> {
 
   hasSettingsChanged = (a, b) => a.settings !== b.settings;
 
-  hasAccountsChanged = (a, b) => a.accounts !== b.accounts;
+  hasAccountsChanged = (a, b) => {
+    return b.accounts.active.some(
+      account =>
+        !isEqual(account, a.accounts.active.find(ab => ab.id === account.id)),
```
gre, line 68: i don't agree with the changes done here. isEqual feels very complex in performance cost. why are we doing this exactly?

LFBarreto: the DB is updated everytime the app updates so we're actually writing in DB at each navigation action or anything that triggers an update on the store we should really move this kind of logic into redux directly imo

gre:
closing in favor of continuation in #1250

PR #1247, opened 2020-04-16
Libcore bindings 4.11.0 + ledgerjs update

gre:
https://github.com/LedgerHQ/ledger-live-desktop/pull/2864


Arnaud97234:
### Bug found
Both iOS and Android. During the send flow step 2 (Recipient address) when user paste or flash  the recipient address the following error is displayed:
"Oops, internet seems down"

### How to reporduce
- Go to Stellar account -> Send
- Paste a valid recipient address

gre:
reported bug was fixed. i'll merge optimistically and we'll confirm later / in the testflight that i'll push later tonight

PR #1250, opened 2020-04-17
LL-2295 Android: fixes app to freeze at boot due to SQLite limit reached (2 MB) – change our db to store per account

gre:
(DB): updated accounts storage saving process to split the data by account key and not grouped

SQLite accounts data is now stored by account keys thus removing certain limitation of storage we can have on android

### Type

Bug Fix

### Context

LL-2295

**Surface:** on iOS and Android, everything related to the app data is touched. Especially and mainly the accounts data that have been optimized and that fixes a limit getting reach when you have high number of accounts loaded.

**Compatibility:** coming from an old app version TO this PR version should properly migrate your accounts in a way that it is totally smoothless for you to see this happening. **However** moving back to old version (of develop) is NOT supported and you need to reset your app, something that can never be done on store as we never go backward.

**New users:** coming from scratch without existing accounts obviously should be correctly working.

**Removing accounts** need to be tested a bit because the code have logic for it to properly work (typically you should never see again the account you have deleted)

**Interruption of the app / Restart of the app** is something that should be tested because it's the way you stop the save() and read() again from the db.  Note that interrupting the app JUST after an action like editing the account IS NOT guaranteed to have been saved but things must not break when you come back and in practice the device is fast enough to have time to save things. (we still have implemented throttled but it's under 1s)

**account ordering** should be preserved.

**clear cache** should be intact.

more info on the main issue

### DIFF
--- a/src/components/DBSave.js
+++ b/src/components/DBSave.js
@@ -1,61 +1,78 @@
 // @flow
-import { Component } from "react";
-import throttle from "lodash/throttle";
-import { connect } from "react-redux";
-import db from "../db";
+import { useRef, useEffect, useMemo } from "react";
+import identity from "lodash/identity";
+import throttleFn from "lodash/throttle";
+import { useSelector } from "react-redux";
 import type { State } from "../reducers";
 
-const instances = [];
-
-export const flushAll = () => Promise.all(instances.map(i => i.save.flush()));
-
-class DBSave extends Component<{
-  dbKey: string,
-  state: State,
-  hasChanged: (next: State, prev: State) => boolean,
-  lense: (s: State) => any,
+type Props<Data, Stats> = {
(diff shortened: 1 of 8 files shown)


juan-cortes:
I'm able to make the app not launch at all still while in this PR. You can easily replicate it by changing the scanned accounts in mock to something like 500 https://github.com/LedgerHQ/ledger-live-common/blob/6b7b26427fa001ad2d32f6ff0dfcdf2a396f4f1e/src/bridge/mockHelpers.js#L132-L138 and reducing the delay. With 400 accounts I crash, this would presumably be the same with fewer accounts but thousands of operations.

image

gre:
not related to the account storage but to JS debugging execution it seems.

PR #1264, opened 2020-05-11
LL-2422 

LFBarreto:
(Manager): glitch on progress bar install issue

While several install and uninstalls are occurring the progress bar now triggers only on the active un/install progression.

### Type

UI Polish

### Context

LL-2422

Manager screen

### DIFF
--- a/src/screens/Manager/AppsList/AppInstallProgress.js
+++ b/src/screens/Manager/AppsList/AppInstallProgress.js
@@ -44,25 +44,21 @@ export const InstallProgress = ({
           />
         </LText>
       </View>
-      {progress !== 1 && installing ? (
-        <ProgressBar
-          progressColor={colors.live}
-          style={styles.progressBar}
-          height={6}
-          progress={progress * 1e2}
-        />
-      ) : (
-        <InfiniteProgressBar
-          progressColor={colors.live}
-          style={styles.progressBar}
-          height={6}
-        />
-      )}
+      <ProgressBar
+        progressColor={colors.live}
+        style={styles.progressBar}
+        height={6}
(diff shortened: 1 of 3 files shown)


src/screens/Manager/AppsList/index.js
```diff
@@ -35,7 +35,7 @@ const AppsList = ({
   setStorageWarning,
   isInstalledView,
 }: Props) => {
-  const viewHeight = active ? "auto" : height - 267;
+  const viewHeight = active ? "auto" : height - 267 || 0;
```
gre, line 38: can you explain what `|| 0` is doing here?

gre: btw with precedence it's useless code. it evals as `active ? "auto" : height - (267 || 0);`

LFBarreto: i'm not sure as to how but while testing the main feature i got a random error on this viewHeight that gave a NaN instead of a number and broke the stylesheet. could not reproduce it easily so i just figured this could help :shrug:

LFBarreto: ho i should fix with parenthesis then

src/screens/Manager/AppsList/AppInstallProgress.js
```diff
@@ -44,25 +44,21 @@ export const InstallProgress = ({
           />
         </LText>
       </View>
-      {progress !== 1 && installing ? (
-        <ProgressBar
-          progressColor={colors.live}
-          style={styles.progressBar}
-          height={6}
-          progress={progress * 1e2}
```
gre, line 52: is 1e2 what the cool kids uses instead of 100? :trollface:

PR #1267, opened 2020-05-11
LL-2447

LFBarreto:
(Analytics): track screen events as pure events with same formatting as LLD

### Type

Analytics

### Context

LL-2447

All screen analytics should now be tracked as events maybe see with @aramab for details as how to test this.


gre:
we'll merge and confirm on analytics side later.

PR #1269, opened 2020-05-13
Hotfix import accounts

JunichiSugiura:
- Fix crash on navigation after importing accounts via QR code.
- Fix warning on long press token account on accounts screen (debug build)
- Fix warning on long press token account on account screen (debug build)


Arnaud97234:
@JunichiSugiura Could you give more context please.

PR #1287, opened 2020-06-03
Fix flow types (partially)

JunichiSugiura:
- Upgrade flow-typed definition files for third-party modules.
- Removed some outdated definition files.
- Add custom type definitions for react-i18next since its flow-typed version is outdated.
- Fixed account related errors in few components. (I'll create another PR to fix them all)


src/families/ripple/ScreenEditFee.js
```diff
@@ -27,7 +28,11 @@ type RouteParams = {
 };
 
 function RippleEditFee({ route }: Props) {
-  const { account } = useSelector(accountScreenSelector(route));
+  const { account: accountLike } = useSelector(accountScreenSelector(route));
+
+  if (!accountLike) return null;
+
+  const account = getMainAccount(accountLike);
```
gre, line 35: it's supposed to be `getMainAccount(accountLike, parentAccount)`
now if there is no case where this can happen, i suggest to just do `invariant(account.type==="Account"...`

src/screens/UnfreezeFunds/01-Amount.js
```diff
@@ -84,9 +88,21 @@ type RouteParams = {
...
 
-export default function UnfreezeAmount({ navigation, route }: Props) {
-  const { account } = useSelector(accountScreenSelector(route));
-  invariant(account, "account is required");
+export default function UnfreezeAmount({ route }: Props) {
+  const { account: accountLike } = useSelector(accountScreenSelector(route));
+  if (!accountLike) {
+    return null;
+  }
+  const account = getMainAccount(accountLike);
```
gre, line 96: same

PR #1295, opened 2020-06-15
Transaction confim fields config

LFBarreto:
Use live-common transactionConfirmFields config in order to show associated operation / families transaction confirmation device step UI.

future live-common config updates should apply automatically unless theres a need for more specific UI.

### Type

Feature

Try operations of each family we support no need to validate the transaction just go to the device validation step.
bitcoin
ethereum
ripple
stellar
tezeos
tron
cosmos


LFBarreto:
all changes merged into #1299

PR #1298, opened 2020-06-17
LL-2695 Fix styling for devicejob error bottom modals

juan-cortes:
Group

Added some padding for the error bottom modal for devicejobs and increased the margin to follow the design guidelines, the descriptions should stop narrower than the button, we tend to forget this.

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-2695

Try to access a device and force a disconnect on the manager, or any other flow that tries to connect to the device. On the screenshots there are two of the options, but there are more cases, all should look good and not like on the linked issue.


juan-cortes:
Rebased this since it's already two weeks old

PR #1299, opened 2020-06-17
Cosmos bugfixes

JunichiSugiura:
Use live-common transactionConfirmFields config in order to show associated operation / families transaction confirmation device step UI.

future live-common config updates should apply automatically unless theres a need for more specific UI.

+ UI fixes mainly for smaller screens
+ bump live-common v12.33.3

### Type

Feature

Try operations of each family we support no need to validate the transaction just go to the device validation step.
bitcoin
ethereum
ripple
stellar
tezeos
tron
cosmos


gre:
we'll continue to test and confirm all is fine with the testflight build. we'll merge this work for now.

PR #1301, opened 2020-06-22
[Cosmos] Add ScrollView to SelectAmount

JunichiSugiura:
### Context

The cta on SelectAmount screen is cut off on smaller devices like iPhone 5s. This PR will squeeze the height of the contents so that user can see all information on that screen. Plus, it adds functionality to click outside the keyboard to hide the keyboard.

- Select Amount screen on delegation/redelegation flow.

### DIFF
--- a/src/families/cosmos/shared/02-SelectAmount.js
+++ b/src/families/cosmos/shared/02-SelectAmount.js
@@ -278,6 +278,7 @@ const styles = StyleSheet.create({
     flex: 1,
     alignItems: "center",
     justifyContent: "center",
+    minHeight: 150,
   },
   inputStyle: { textAlign: "center", fontSize: 40, fontWeight: "600" },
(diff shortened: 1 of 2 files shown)


src/families/cosmos/shared/02-SelectAmount.js
```diff
@@ -257,7 +278,6 @@ const styles = StyleSheet.create({
     flex: 1,
     alignItems: "center",
     justifyContent: "center",
-    minHeight: 150,
```
LFBarreto, line 260: Leave this one here without that the whole screen collapses on itself on android :sweat_smile:

JunichiSugiura: ah, that makes sense

LFBarreto:
closing this i think the issue was fixed already

PR #1304, opened 2020-06-22
LL-2544 NativeSeg/Seg

LFBarreto:
Select the first scanned account on the add accounts list by default same way as its done on LLD

HODL: needs a live-common bump with the new derivation order

### Type

Feature

### Context

LL-2544

Try add a new bitcoin account the native segwit should be selected once it appears on the list.


gre:
still needed even with https://github.com/LedgerHQ/ledger-live-mobile/pull/1319 ? to be tested

LFBarreto:
will be included in #1319

PR #1320, opened 2020-07-08
LL-2757 Carousel

juan-cortes:
~This is a work in progress of the carousel feature for mobile. I never created the PR because I'm not happy with the performance and feel when running on a real device, it seemed sketchy. I was using react-navigation slides and thought that might have been the issue so I tried a dedicated library for it, no improvement whatsoever.~

~If someone else takes it, all that remains apart from perf, is not showing the carousel if we are on the placeholder mode on the porfolio.~ Thank you @LFBarreto for your help

The same feature as on desktop, make sure the carousel shows and links to the right places.
It shouldn't show if we are showing an empty state placeholder for the portfolio.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2757


juan-cortes:
Might want to rethink this here too due to the perceived backlash from users @gre 
I believe @Arnaud97234 error is unrelated to the feature but still I'd HODL this until we have a redesign or drop it completely.

juan-cortes:
Closing this until discussed

PR #1331, opened 2020-07-24
LL-2775 (Fonts): updated fonts to Inter family

LFBarreto:
Updated fonts to Inter family.
everything should work fine just need to be double checked on IOS as i could not build it myself and test on it.

### Type

UI Polish

### Context

LL-2775

All fonts


nabil-brn:
Still not building, I'll ask to a dev with a mac  to fix this issue.

juan-cortes:
I can check

juan-cortes:
According to https://github.com/oblador/react-native-vector-icons/issues/1074 there's a bug where we end up with double-copy resources, in this case the fonts. I followed what they say in that thread about removing the duplicated resources and it seems to work, I just tested building on the iphone 6 from ledger.

I committed a fix.

@LFBarreto @nabil-brn

khalilbenihoud:
Tested on a Google PIXEL 3a on my side, and it's top notch.

PR #1334, opened 2020-07-28
live-common 13.5.0

gre:
- live-common bump: please see full detail in https://github.com/LedgerHQ/ledger-live-common/releases/tag/v13.5.0
- some minor libs bump. check integrity of the app on android and ios should be enough (not too deep).


nabil-brn:
I'm not able to reproduce the issue on TRC20, apparently trongrid fixed the issue, so we no longer need those changes.

PR #1337, opened 2020-07-30
upgrade dependencies 

LFBarreto:
upgrade dependencies mainly react-native to ~0.63.2~ 0.63.3

react-native-ble also needed upgrade

### Type

Upgrade

Just build the app everything should work fine.

Build it on staging and test the bluetooth as well


gre:
awaiting conflict resolution

gre:
conflicts :'(

package.json
```diff
@@ -110,10 +110,10 @@
     "re-reselect": "^3.4.0",
     "react": "^16.13.1",
     "react-i18next": "11.7.0",
-    "react-native": "0.61.2",
+    "react-native": "0.63.2",
```
MortalKastor, line 113: This is not obvious because React Native uses the minor part of their version as a major, introducing invasive changes, and nothing really blocks you from doing so, but you can't just bump RN minors as any other package.
You need to check what changes had been made for everything React Native outside of `node_modules` and apply them to our repo. Thankfully they now provide a handy upgrade helper, listing them all from any specific version to any other : https://react-native-community.github.io/upgrade-helper/?from=0.61.2&to=0.63.2

At a glance, you'll see that 20+ files changed from our RN version to the one you're targeting, versus only 8 in this PR, so we're missing stuff.

LFBarreto: Yes i've seen that one i just didn't need to change anything on android was working as is, so i skiped on the IOS part my bad :sweat_smile: I'll try and look for a solution with someone that can build on IOS since i can't.

JunichiSugiura:
> yarn pod fails with the following error:
> 
> ```
> [!] CocoaPods could not find compatible versions for pod "ReactCommon/jscallinvoker":
>   In snapshot (Podfile.lock):
>     ReactCommon/jscallinvoker (from `../node_modules/react-native/ReactCommon`)
> 
>   In Podfile:
>     ReactCommon/jscallinvoker (from `../node_modules/react-native/ReactCommon`)
> 
> None of your spec sources contain a spec satisfying the dependency: `ReactCommon/jscallinvoker (from `../node_modules/react-native/ReactCommon`)`.
> ```

Yup, I experienced the same yesterday. I'll take a deeper look tomorrow.

MortalKastor:
The React Native upgrade should be good on iOS now (poke @JunichiSugiura). I did a bit of polish on the Android front too (so you probably want to test again on that platform @Arnaud97234).
Finally, I rebased on the current develop and force pushed (sorry for the mess I did in the discussion/history).

gre (dismissed):

gre:
can we merge develop and potentially squash to one commit?

valpinkman (approved):

Arnaud97234 (changes requested):
###Bug found
Android:
1) App crashes just after importing TRON accounts
image
image (1)

2) ATOM - Add accounts:
Scanning for account keeps looping without discovering any accounts

valpinkman:
@LFBarreto told me that the issue would be also on the develop as the explorers/countervalues could return NaN values and thus make the graph breaks

gre:
ok we will deal with the NaN problem independently of this PR. do I assume this was the only issue @Arnaud97234 ❓

Arnaud97234 (approved):

PR #1342, opened 2020-08-06
add error and CTA to buy ETH when NotEnoughGas

valpinkman:
Provide a way to go to Buy whenever parent ETH account does not have enough funds to pay for the fees of the ERC20 transaction

### Type

Feat

### Context

LL-2684

Send Summary

IMG_1E97102EB118-1

### DIFF
--- a/src/screens/SendFunds/04-Summary.js
+++ b/src/screens/SendFunds/04-Summary.js
@@ -114,7 +114,8 @@ export default function SendSummary({ navigation, route }: Props) {
 
   if (!account || !transaction || !transaction.recipient) return null; // FIXME why is recipient sometimes empty?
   const { amount, totalSpent, errors } = status;
-  const { transaction: transactionError, gasPrice } = errors;
+  const { transaction: transactionError } = errors;
+  const error = status.errors[Object.keys(status.errors)[0]];
   const mainAccount = getMainAccount(account, parentAccount);
 
   // console.log({ transaction, status, bridgePending });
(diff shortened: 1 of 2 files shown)


src/screens/SendFunds/04-Summary.js
```diff
@@ -140,6 +148,16 @@ export default function SendSummary({ navigation, route }: Props) {
           transaction={transaction}
           navigation={navigation}
         />
+        {gasPrice && gasPrice instanceof NotEnoughGas ? (
```
gre, line 151: mmh IMO we shouldn't do this specific code.

we should use

```
error = status.errors[Object.keys(status.errors)[0]]
```

because any error is a case of blocker and we should show it to user (show the first one at least)

valpinkman: This is a new block specifically made for this error, I have not touch the rest of the UI

gre: mmh to be confirmed but imo we want to do such thing whatever the error is. for instance, we have same issue on XRP if you go increase the fees from that screen. which would not be caught here.

valpinkman: I still don't know what do with this then ? @gre @cthepot-ledger @khalilbenihoud

gre: even if this could be enough as-is, I still think it would be better to generalize this to the equivalent of the logic of the NEXT that I think with this idea of taking first error in status.errors

this would address it more generically like for XRP and to me this is compatible with the idea of any error as here you are already using TranslatedError

gre: to summarize:
in the Summary step, there is rarely a case where an error exist, like 99% of the time, status.errors is settled to {} (no errors). but when it happens, we want that error to be displayed.

valpinkman: 👍

valpinkman:
Ready to be reviewed

Arnaud97234 (approved):

Arnaud97234 (changes requested):
Discussed with @aramab : Could we add a Cancel button below / above the Buy ?

gre:
@Arnaud97234 please add another task for it. it's out of scope of the initial LL task.

PR #1347, opened 2020-08-13
use erc20 unit when available in validation screen

valpinkman:
### Type

Fix

### Context

LL-3055

Send Flow Validation Screen

### DIFF
--- a/src/components/ValidateOnDevice.js
+++ b/src/components/ValidateOnDevice.js
@@ -45,7 +45,7 @@ function AmountField({
 }: FieldComponentProps) {
   let unit;
   if (account.type === "TokenAccount") {
-    unit = account.token.units[0];
+    unit = getAccountUnit(account);
   } else {
     const mainAccount = getMainAccount(account, parentAccount);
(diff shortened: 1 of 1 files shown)


src/components/ValidateOnDevice.js
```diff
@@ -43,8 +43,13 @@ function AmountField({
...
 }: FieldComponentProps) {
-  const mainAccount = getMainAccount(account, parentAccount);
-  const unit = getAccountUnit(mainAccount);
+  let unit;
+  if (account.type === "TokenAccount") {
+    unit = account.token.units[0];
+  } else {
+    const mainAccount = getMainAccount(account, parentAccount);
+    unit = getAccountUnit(mainAccount);
+  }
```
gre, line 52: mmh isn't it all that just like

```
const unit = getAccountUnit(account)
```
?

valpinkman: will fix then

gre:
we can merge this work for now, can be refined later

PR #1353, opened 2020-08-19
(Fonts): replace Rubik font

LFBarreto:
Remove deprecated font Rubik from styles

### Type

Bug Fix

Send flow currency inputs should now be inter like everything else


valpinkman (approved):

gre (changes requested):
Mmh we never should use the "default" font but explicitly tell which font to use

IAmMorrow (approved):

PR #1354, opened 2020-08-20
added warning message on token send flow when user doesn't have enough ETH in parent acc…

IAmMorrow:
…ount to pay for TX fees

### DIFF
--- a/src/components/SendRowsFee.js
+++ b/src/components/SendRowsFee.js
@@ -9,11 +9,13 @@ export default ({
   account,
   parentAccount,
   navigation,
+  status,
 }: {
   transaction: any,
   account: AccountLike,
   parentAccount: ?Account,
   navigation: any,
+  status: TransactionStatus,
 }) => {
   const mainAccount = getMainAccount(account, parentAccount);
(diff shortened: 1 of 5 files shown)


Arnaud97234 (changes requested):
In case of parent balance too low we should disable the continue button in send flow step summary.

IMG_DD5BDBFC2792-1

gre (approved):

gre:
Mmh what is this PR doing vs https://github.com/LedgerHQ/ledger-live-mobile/pull/1342 ?

I kinda prefer https://github.com/LedgerHQ/ledger-live-mobile/pull/1342 because it took the generalisation a bit farther (it renders ANY error that would remain in the status.errors, not specific to eth), but why do we have 2 similar ticket in the first place?

Arnaud97234:
> Mmh what is this PR doing vs #1342 ?
> 
> I kinda prefer #1342 because it took the generalisation a bit farther (it renders ANY error that would remain in the status.errors, not specific to eth), but why do we have 2 similar ticket in the first place?

@gre  Yes actually they are addressing the same topic. I guess we could close that one and keep the one from Val.

PR #1359, opened 2020-08-25
bump libcore bindings 4.15.10

LFBarreto:
bump libcore bindings should fix cosmos amplifier issue


Arnaud97234 (changes requested):
### Bug found
The following error is occurring on **Android** during the send flow (Send, Undelegate, Delegate...). I can see the error in undelegate flow after selecting the amount to undelegate when I land on device selection screen.
The error will prompt also when sending funds -> Step Amount.
ezgif com-video-to-gif

Arnaud97234 (approved):
All cosmos tx are OK, other coins as well

nabil-brn (approved):

PR #1372, opened 2020-08-28
add new format to firmware hash

valpinkman:
LL-3066
split FW hash to be displayed the same way nano S|X display them

We do not have the firmware update implemented on mobile yet, but as a safe guard, updating the component so it doesn't break

https://github.com/LedgerHQ/ledger-live-common/pull/842 waiting for it to be merged and live-common bumped


LFBarreto (approved):

gre:
merged because no impact on this as the firmware update is not activated in LLM. but it is important for incoming PR on live-common

PR #1389, opened 2020-09-24
fix tokens unit on validate on device

LFBarreto:
(ValidateOnDevice): fix edge case where amount unit was set to parent unit instead of tx subAccount one

### Type

Bug Fix

### Context

Slack

Try sending tokens (ERC20...) the amount on device validation step should be in the correct unit now


LFBarreto:
closing this for #1390 is a better solution to fix the issue

PR #1393, opened 2020-09-29
LL-3116 Use the 'clearAccount' helper for CLEAN_CACHE

juan-cortes:
We were not making use of the helper from live-common, resulting in a partial clear cache since subAccounts were not affected. The linked Jira tasks says erc20 were being flushed but I don't see how that's possible. As a side note, on a few occasions I had to clear cache **twice** in order to see this "No operations" screen, but this can't be related to this bug, and is perhaps a good thing to take a look into as a separate bug.

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-3116

- Import Tron, Algorand, and Ethereum accounts with subAccounts
- Turn OFF network connection
- Clear the application cache
- Go to Portfolio page

### DIFF
--- a/src/reducers/accounts.js
+++ b/src/reducers/accounts.js
@@ -11,6 +11,7 @@ import {
   importAccountsReduce,
   isUpToDateAccount,
   withoutToken,
+  clearAccount,
 } from "@ledgerhq/live-common/lib/account";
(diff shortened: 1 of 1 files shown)


gre:
Woo good catch

gre (approved):

PR #1394, opened 2020-10-02
Add analytics events for LLM swap

juan-cortes:
Adds analytics events tracking for the swap flow just like we did for LLD.

### Type

Feature

Test that the swap flow doesn't break due to this, it shouldn't, but hey, worth checking.


gre (approved):
safe to merge

PR #1397, opened 2020-10-06
LL-3570 Swap - Auto update swap status on history/op details

juan-cortes:
Similar to what we did for LLD, introduce an auto update for swap operation status that will update both the History and the operation details if we are currently viewing a particular swap. This is done by polling the endpoint every 10 seconds until the status reaches an end (finished, canceled, refunded).

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3570

- Either manually modify the app state to have a pending swap, or create a new swap.
- Go into the swap history
- Go into the swap details for that swap
- Wait until the swap status changes to finished.

I kept the pull to refresh on the history screen which manually triggers the update but it doesn't affect the automatic polling.

### DIFF
--- a/src/screens/Swap/History/index.js
+++ b/src/screens/Swap/History/index.js
@@ -1,4 +1,10 @@
-import React, { useRef, useEffect, useState } from "react";
+import React, {
+  useCallback,
+  useRef,
+  useEffect,
+  useState,
+  useMemo,
+} from "react";
 import {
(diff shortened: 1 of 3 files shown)


juan-cortes:
Closed in favour of https://github.com/LedgerHQ/ledger-live-mobile/pull/1401

PR #1403, opened 2020-10-07
LL-2908 (AddAccounts): conditional token cta updated) 

LFBarreto:
(AddAccounts): conditional token cta updated) 

Screenshot_2020-10-07-12-20-14-400_com ledger live debug

Screenshot_2020-10-07-12-20-53-496_com ledger live debug

Screenshot_2020-10-22-13-46-06-123_com ledger live debug

### Type

Feature
### Context

LL-2908

Add accounts with tokens, check JIRA task for the different conditions of the CTA


gre:
HODL for discussion on LLD https://github.com/LedgerHQ/ledger-live-desktop/pull/3218

gre (approved):

Arnaud97234 (dismissed):

juan-cortes:
I believe this https://ledgerhq.atlassian.net/browse/LL-3669 task is only happening on this particular PR. Instead of passing the currency as an initial query we should filter the accounts by currency as done buy/swap and leave the input empty cc @LFBarreto

src/screens/AddAccounts/02-TokenCurrencyDisclaimer.js
```diff
@@ -64,13 +68,48 @@ export default function AddAccountsTokenCurrencyDisclaimer({
...
+
+  // specific cta in case of token accounts
+  const onTokenCta = useCallback(() => {
+    if (parentTokenAccount && parentTokenAccount.type === "Account") {
+      onClose();
+      navigation.navigate(NavigatorName.ReceiveFunds, {
+        screen: ScreenName.ReceiveSelectAccount,
+        params: {
+          // prefill search input with account name or token name
+          currency: tokenAccount ? token.name : parentTokenAccount.name,
```
juan-cortes, line 96: Filter the accounts in https://github.com/LedgerHQ/ledger-live-mobile/blob/da9111ce18461c0f60b58ca5a8eff7df609359e6/src/screens/ReceiveFunds/01-SelectAccount.js#L30 if we receive a currency, instead of passing the currency as an initial query for the input search as per https://ledgerhq.atlassian.net/browse/LL-3669

juan-cortes (changes requested):

src/screens/AddAccounts/02-TokenCurrencyDisclaimer.js
LFBarreto: I updated it accordingly :+1:  as we discused before with @Arnaud97234  the main issue was to do a similar UX as swap in the filtering and prefilling the sub accounts with the selectedCurrency

Arnaud97234 (changes requested):
The only issue remaining is for the following use case:
- An ETH account is present but there is no subAccount corresponding to the selected ERC20.
https://ledgerhq.atlassian.net/browse/LL-3669

Arnaud97234 (approved):

gre (approved):

PR #1406, opened 2020-10-07
LL-3363 Fix for experimental device action being active by default

juan-cortes:
Introduces the experimental device flow for opening applications (prompting the user to open) on the device. It takes forever to work though so perhaps we should consider improving the performance. According to my testing it's somewhere between 16 and 18 seconds between the user accepting on the device and the import continuing, which is huge.

### Type
Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-3363

- Launch LLM
- Go into a flow the requires to be on a currency app
- Test the behaviour while being on the dashboard/inside another app/etc


gre (dismissed):

Arnaud97234 (changes requested):
### Bug found
When a currency app is missing on the device and user go to Add Account flow for that currency:

juan-cortes:
I'm checking @Arnaud97234

juan-cortes:
Good catch, something like https://github.com/LedgerHQ/ledger-live-common/pull/906 should be merged on live-common in order to prevent this

gre:
@juan-cortes tomorrow you can release a new livecommon , you will need to bump all @ledgerhq/ libs

gre (approved):

Arnaud97234 (approved):

PR #1414, opened 2020-10-16
Add BCH hardfork warning banner

gre:
Like https://github.com/LedgerHQ/ledger-live-desktop/pull/3238 but only on portfolio.

It should look like figma design with no external link icon as per design and since i could not get it to align properly with the rest of the text.

Screenshot_2020-10-20-10-33-55-130_com ledger live debug


Arnaud97234 (changes requested):
### Bug found

On portfolio, "learn more" text is missing and the design is broken in the Warning displayed at the top.
IMG_CCB55FC013FD-1

Arnaud97234 (approved):

PR #1422, opened 2020-10-21
LL-3335 Compound

LFBarreto:
Compound

### Type

Feature
### Context

LL-3335


src/screens/Lending/useCompoundSummaries.js
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    const newSummaries = makeSummaries(accounts);
+    if (newSummaries) {
+      setSummaries(newSummaries);
+    }
+  }, [accounts, setSummaries]);
+
+  return summaries;
+}
```
gre, line 41: can we unify these back into live-common and also consider merging with https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/screens/lend/useCompoundSummaries.js ? no reason to duplicate these.

LFBarreto: https://github.com/LedgerHQ/ledger-live-common/pull/936 :+1:  i'll update the code once this is merged

PR #1423, opened 2020-10-21
LL-2757 - Promo carousel

juan-cortes:
Here's the toned down version of the original carousel with less intrusive UI and smaller footprint. The ability to close the banner and not see it again until we bump the nonce with new slides so that the user doesn't miss new features/announcements. And of course the ability to bring it back (shall they want to do this) from the General settings screen.

slides

This is based on the work from https://github.com/LedgerHQ/ledger-live-mobile/pull/1320, including the performance gains achieved by @LFBarreto https://github.com/LedgerHQ/ledger-live-mobile/pull/1320/commits/7ecd701b27ef5bd4c889778aec884d99a77a865d on the carousel and a fix to the bug https://github.com/LedgerHQ/ledger-live-mobile/pull/1320#pullrequestreview-452779847 described here that @Arnaud97234 found, so while testing this, also make sure the scroll to top feature is not broken.

It includes the placeholders for the slides for Vote and Sell, the dev that implements those feature just has to enable the slide and link it to the right place when the time comes. Also I added the Algorand slide (cc @cthepot-ledger or @aramab) that wasn't listed in the ticket but that is shown on desktop so it only makes sense to add it here to.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2757

- Launch the app
- Test that each individual slide links to the appropriate section/url
- Test the dismiss feature
- Bring the carousel back from the settings
- Test that it doesn't come back on relaunch


gre:
@JunichiSugiura can you take the code review?

Arnaud97234 (approved):

gre (approved):

PR #1424, opened 2020-10-22
LL-3679 Make settings/accounts navigators go to first screen on blur

juan-cortes:
When the user navigates to a secondary/deeper screen from the bottom tab navigator and then navigates to another tab, reset the first tab to the initial screen so that when we come back it's not still in the secondary screen. Was that clear enough? I don't think so, I'll explain better in the test plan.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3679

- Press the settings tab
- Click on General
- Click on the Accounts tab
- Click on the Settings tab.

We should now be on the main settings screen and **not** in the _General_ one. The same thing applies to the Accounts navigator, you should be able to access a particular account screen, go to the portfolio, come back to the accounts tab and be on the main screen and not the nested one.

### DIFF
--- a/src/components/RootNavigator/MainNavigator.js
+++ b/src/components/RootNavigator/MainNavigator.js
@@ -34,6 +34,7 @@ export default function MainNavigator() {
         name={NavigatorName.Accounts}
         component={AccountsNavigator}
         options={{
+          unmountOnBlur: true,
           tabBarIcon: (props: any) => (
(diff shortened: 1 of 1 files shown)


LFBarreto (approved):

juan-cortes:
cc @nabil-brn

Arnaud97234 (approved):

PR #1437, opened 2020-11-03
Ll 3800 Build apk on every commit

machard:
add android build on CI workflow

### Type

Improvement

### Context

https://ledgerhq.atlassian.net/browse/LL-3800

check if there is an android apk available in this commit !

### DIFF
--- a/.github/workflows/bundle-app.yml
+++ b/.github/workflows/bundle-app.yml
@@ -1,0 +1,71 @@
+name: bundle app
+on: [push, pull_request]
+
+jobs:
+  bundle-apk:
+    runs-on: macos-latest
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.0
+        with:
+          access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
+      - name: install bundle
+        run: |
+          gem install bundler:2.1.4
+      - uses: ledgerhq/actions/check-member@v2.0.0
+        with:
+          username: ${{ github.actor }}
+          ban: ledgerlive
+        if: github.event_name == 'pull_request'
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@main
+        with:
+          node-version: 12.x
+      - name: set git user
+        run: |
+          git config user.email "team-live@ledger.fr"
+          git config user.name "Team Live"
+      - name: install dependencies
+        run: yarn --frozen-lockfile
+      - uses: ledgerhq/actions/get-package-infos@v1.0.0
+        id: version
+      - name: set beta name
+        uses: ./.github/actions/change-product-name
+      - name: Get short SHA
+        id: slug
+        run: echo "::set-output name=sha8::${GITHUB_SHA:0:7}"
+      - name: make local version
+        run: yarn version --new-version=${{ steps.version.outputs.clean }}-sha.${{ steps.slug.outputs.sha8 }}
+      - name: build the app
+        run: |
+          export ANDROID_KEYSTORE_FILE=$(pwd)/android/app/staging.kstr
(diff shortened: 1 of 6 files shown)


juan-cortes (dismissed):
lgtm, perhaps we could skip the uploads of the less used versions, arm64-v8a represents 95% of our users.

.github/workflows/bundle-app.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  bundle-apk:
+    runs-on: macos-latest
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.0
+        with:
+          access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Ruby
```
JunichiSugiura, line 12: I believe Ruby/bundler isn't necessary for Android build.
Either changing OS to ubuntu or add special flag to avoid bundle install here to save some time.
https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/scripts/post.sh#L28

.github/workflows/bundle-app.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: echo "::set-output name=sha8::${GITHUB_SHA:0:7}"
+      - name: make local version
+        run: yarn version --new-version=${{ steps.version.outputs.clean }}-sha.${{ steps.slug.outputs.sha8 }}
+      - name: build the app
+        run: |
+          export ANDROID_KEYSTORE_FILE=$(pwd)/android/app/staging.kstr
+          export ANDROID_KEYSTORE_PASS=staging
+          export ANDROID_KEY_ALIAS=staging
+          export ANDROID_KEY_PASS=staging
+          bundle exec fastlane android apk --env production
```
JunichiSugiura, line 49: I'm not sure `--env production` flag is used inside fastlane.
We can replace this line with `yarn android:apk` alias

.github/workflows/bundle-app.yml
JunichiSugiura: Like skip if "${GITHUB_WORKFLOW}" == "bundle app"

machard: I switched from ubuntu to ios because of a "System limit for number of file watchers reached" error.

Regarding the ruby bundle thing, I squashed all commits, but I recall I had to execute bundle install.

That's also why I did https://github.com/LedgerHQ/ledger-live-mobile/pull/1437/files/2ee883b9494405fcad5876bce0dde28f5b1c7aec#diff-c7a2bbcab4d0757f7a7ba519bf73e7af3dbdb641ff4fdee2b70b684586f49cbdL25.
However it looks like I made this change when I was trying with ubuntu

.github/workflows/bundle-app.yml
machard: let's try

.github/workflows/bundle-app.yml
machard: the pod install could be skip though

machard: However you will need it right ?

https://github.com/LedgerHQ/ledger-live-mobile/pull/1437/files/2ee883b9494405fcad5876bce0dde28f5b1c7aec#diff-c7a2bbcab4d0757f7a7ba519bf73e7af3dbdb641ff4fdee2b70b684586f49cbdL25 => I can revert that since we are using macOS but if I did it i am 99% sure it blocked when I was trying with ubuntu

machard: I could also switch back to Ubuntu with https://github.com/expo/expo-github-action/issues/20#issuecomment-541676895

gre (approved):
I think the work here is good enough. We can polish later but I'll merge to benefit having the builds in incoming PRs 💪

.github/workflows/bundle-app.yml
JunichiSugiura: oh yeah, I forgot that we're using fastlane 😅  sorry!

PR #1441, opened 2020-11-04
LL-3609 Better flow when accessing manager from Swap with query/tab

juan-cortes:
### Original issue in the ticket
When following an error flow end from the swap (will likely need to be integrated on Sell cc @IAmMorrow) there is a possibility that the user can have a missing exchange app, an outdated exchange app, or any number of missing/outdated currency apps. In those cases, we will prompt the user to go to the manager and install or update the app accordingly.

The first iteration of this merely redirected the user to the manager but was further improved by the work of @LFBarreto introducing the `searchQuery` concept that we could pass as a route param to the navigator and have the screen pre-filled to that currency's name. This was great, but being the nonconformist beings we are we decided it would be nicer to preselect the installed tab and show the banner to update, if that was the action the user was coming to the manager for after-all.

Cool, cool. We introduce a new flag to show the update modal, change a bit the logic depending on missing/outdated, pass a default tab for the manager to be able to be on the installed tab by default, and test it all works.

### Uncovered issue found while doing the ticket
After accessing the manager with a pre-filled search query, **any future access to the manager will include that search query.** So if the user didn't have the exchange app installed, accessed the manager from the swap flow, installed the app and then did other things... When that user accessed the manager again during this same session, it will once again show a filtered catalog with the "exchange" app only on that list.

**Why?** Because that's the expected behaviour from https://github.com/react-navigation/react-navigation/issues/6674 and we are expected to reset the route parameters ourselves. (┛ಠ_ಠ)┛彡┻━┻ In order to do this, we need to listen to the `tabPress` event on the navigator definition, but the signature of that parameter which allows us to access the navigator itself required a library bump. Which, luckily, seems to not have broken anything.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3609

### Flow diagram
Group

#### Case 1. Missing exchange app
- [x] Uninstall if needed the exchange app
- [x] Access the swap feature
- [x] Click on the access manager button to install the exchange app
- [x] It is expected that the manager will be in the catalog tab search with the list filtered to only see "exchange"
- [x] Go to the portfolio
- [x] Access the manager from the bottom tab, it is expected that the manager shows the catalog with no filters

#### Case 2. Outdated exchange app
- [x] Install the exchange app, and go into provider 4 to simulate an outdated exchange app
- [x] Access the swap feature
- [x] Click on the access manager button to update the exchange app
- [x] It is expected that the manager will be in the "Installed apps" tab, and the update bottom modal will be visible, allowing you to update the outdated apps.
- [x] Go to the portfolio
- [x] Access the manager from the bottom tab, it is expected that the manager shows the catalog with no filters

#### Case 3. Missing currency app
- [x] Access the swap feature with the exchange app installed and updated
- [x] On the swap form, select a currency that you don't have installed.
- [x] On the bottom modal that shows, click on "Go to Manager" to install the missing currency app
- [x] It is expected that the manager will be in the catalog search with the list filtered to only see the expected currency app or any app that build up on its name. For example, if we are missing Ethereum, the filter will also show Ethereum Classic.
- [x] Go to the portfolio
- [x] Access the manager from the bottom tab, it is expected that the manager shows the catalog with no filters

#### Case 4. Outdated currency app
- [x] Access the swap feature with the exchange app installed and updated
- [x] On the swap form, select a currency that you do have installed but is outdated (provider 4 once again for ETH)
- [x] On the bottom modal that shows, click on "Go to Manager" to update the currency app
- [x] It is expected that the manager will be in the "Installed apps" tab, and the update bottom modal will be visible, allowing you to update the outdated apps.
- [x] Go to the portfolio
- [x] Access the manager from the bottom tab, it is expected that the manager shows the catalog with no filters

Apart from these explicit tests, make sure that the update of react-navigation didn't break anything. Play with the app a bit, I didn't have to make any code changes and I haven't noticed any new crashes or weird behaviour but you never know with this library.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -600,7 +600,7 @@ SPEC CHECKSUMS:
   CocoaAsyncSocket: 694058e7c0ed05a9e217d1b3c7ded962f4180845
   CocoaLibEvent: 2fab71b8bd46dd33ddb959f7928ec5909f838e3f
   djinni_objc: f04ed3e34aeabad13c48893d48cd72c9e2275dbd
-  DoubleConversion: cde416483dac037923206447da6e1454df403714
+  DoubleConversion: 5805e889d232975c086db112ece9ed034df7a0b2
   FBLazyVector: 878b59e31113e289e275165efbe4b54fa614d43d
   FBReactNativeSpec: 7da9338acfb98d4ef9e5536805a0704572d33c2f
   Flipper: be611d4b742d8c87fbae2ca5f44603a02539e365
@@ -611,7 +611,7 @@ SPEC CHECKSUMS:
   Flipper-RSocket: 64e7431a55835eb953b0bf984ef3b90ae9fdddd7
   FlipperKit: ab353d41aea8aae2ea6daaf813e67496642f3d7d
   Folly: b73c3869541e86821df3c387eb0af5f65addfab4
-  glog: 40a13f7840415b9a77023fbcae0f1e6f43192af3
+  glog: 1f3da668190260b06b429bb211bfbee5cd790c28
   ledger-core-objc: 2b7223951d4bcdb999b1c5c75e2359169031e2ba
   lottie-ios: 3a3758ef5a008e762faec9c9d50a39842f26d124
   lottie-react-native: 2a1a82bb326ae51331a5520de0cf706733c6db69
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -600,7 +600,7 @@ SPEC CHECKSUMS:
   CocoaAsyncSocket: 694058e7c0ed05a9e217d1b3c7ded962f4180845
   CocoaLibEvent: 2fab71b8bd46dd33ddb959f7928ec5909f838e3f
   djinni_objc: f04ed3e34aeabad13c48893d48cd72c9e2275dbd
-  DoubleConversion: 5805e889d232975c086db112ece9ed034df7a0b2
+  DoubleConversion: cde416483dac037923206447da6e1454df403714
   FBLazyVector: 878b59e31113e289e275165efbe4b54fa614d43d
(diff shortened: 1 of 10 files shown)


nabil-brn (dismissed):
It's missing the pre filled app name in the catalog for 2 cases.

juan-cortes:
I will take a look @nabil-brn thank you

Arnaud97234 (changes requested):
The behaviour is ok on Android, but on iOS When accessing manager the search input is not preffiled

juan-cortes:
I will check with iOS then, thanks @Arnaud97234

juan-cortes:
tldr @Arnaud97234 you can test again, this should be ok for both platforms now. Sorry for the delay.

-----

Chained modals in react-native-modal are problematic. If we get into a case where we try to show one while the previous one is still visible we can run into problems. In this particular case, the manager's select device modal while entering the feature and the search modal once inside of it were clashing.

The reason for this is that the `onResult` callback of the device action was unsetting `device` from the component's state and as a callback to `setState` we were navigating away, disregarding the modal's [`isOpened`](https://github.com/LedgerHQ/ledger-live-mobile/commit/140fb2e3e8b0a42d3f7b79aa7995a50d4fb04394#diff-f9b1fe378305ecd17d41338add542c6eb89e9feb0b9b9ba578d8b8770517472aR32 ) prop and unmounting from the parent while navigating away.

This meant on iOS that by the time we reached the app list of the manager, as far as iOS/React was concerned, the previous modal wasn't completely hidden, and trying to show a new one resulted in the modal breaking and not rendering at all. On Android it works, although this chained modal issue seems to be recurrent with other wording.

Hacking it to delay the showing of the search modal fixes it because any amount waited means we try to _show_ after the previous one is hidden. By splitting the logic into setting the result and navigating when the modal is closed it all works without this hack.

Arnaud97234 (approved):

Issue #1040, opened 2019-09-13
WalletConnect Integration


ukstv:
- [x] I have checked this feature was not yet requested.

Ledger Live Mobile would be a wonderful mobile Ethereum wallet if it would make using Ethereum accounts on the web. One standard that connects mobile Ethereum wallets to desktop web is [WalletConnect](https://walletconnect.org). The request is integration of WalletConnect into Ledger Live Mobile application.

#### Part of the application

The feature would significantly affect user interface of the application.
1. It would have an additional button for scanning WalletConnect QR codes.
2. It would handle approving Ethereum actions on Nano X dongle through mobile UI.

reuptaken:
While WalletConnect is a way to connect desktop dapps to mobile wallets this could also help with dapps opened in browser on mobile (perhaps not via scanning but taking screenshot of QR code and opening it in Ledger Live. Without this we're unable to interact with Ethereum smart contracts using Ledger Nano X on iOS.

craben20:
Yes, this is a huge user experience flaw. Ledger Live supports ERC-20 storage but doesn't offer a wc:// interface for dApps...

hapm:
I would love to see WalletConnect implemented as well, with qr code and wc:// support.

PR #1452, opened 2020-11-13
Feature/walletconnect

machard:
Fixes https://github.com/LedgerHQ/ledger-live-mobile/issues/1040

previews:
connexion + send transaction : https://drive.google.com/file/d/13J0g_z9TXIWUm_dNM62BsFoH0b0cnO49/view?usp=sharing
sign message : https://drive.google.com/file/d/1tnJ1QLqFhnxDb0iwcYayLV2uobo6v5_s/view?usp=sharing

- [x] connexion management
- [x] deeplinking ios (ledgerlive://wc?uri=WC_URI [wc:...])
- [x] deeplinking android (WC_URI [wc:......])
- [x] send transaction flow (reusing SendFunds flow)
- [x] sign message flow
- [x] sign transaction flow (which dapp to test?)
- [x] broadcast transaction flow (which dapp to test?)
- [x] tracking
- [x] UI polish

I reused the existing QRCode/Camera logic from SelectRecipient, so it can be good to check that i did not break anything. I am not sure the logic to handle missing permissions is still up to date.

test ios deeplinking: url `ledgerlive://wc?uri=pouetpouetpouet`
This uri param is not correct ofc. However, you can get a correct URI from https://example.walletconnect.org/ and then URLencode it.
test android deeplinking: go to https://example.walletconnect.org/


src/screens/SignMessage/03-ConnectDevice.js
```diff
@@ -0,0 +1,146 @@
...
+import DeviceAction from "../../components/DeviceAction";
+import { TrackScreen } from "../../analytics";
+import { ScreenName } from "../../const";
+
+const initialState = {
+  signMessageRequested: null,
+  signMessageError: null,
+  signMessageResult: null,
+};
+const createAction = connectAppExec => {
```
gre, line 24: should this be a new "action" to be put in live-common?

src/screens/WalletConnect/Provider.js
```diff
@@ -0,0 +1,280 @@
...
+        session.session
+          ? {
+              session: session.session,
+            }
+          : {
+              // Required
+              uri,
+              // Required
+              clientMeta: {
+                description: "LedgerLive CLI",
```
gre, line 49: not CLI here.

src/screens/WalletConnect/Provider.js
```diff
@@ -0,0 +1,280 @@
...
+                url: "https://ledger.fr",
+                icons: [
+                  "https://avatars0.githubusercontent.com/u/9784193?s=400&v=4",
```
gre, line 52: probably an icon of Ledger Live is more appropriated.

src/screens/WalletConnect/Provider.js
```diff
@@ -0,0 +1,280 @@
...
+                ],
+                name: "LedgerLive CLI",
```
gre, line 54: just `Ledger Live`

src/screens/WalletConnect/Provider.js
```diff
@@ -0,0 +1,280 @@
...
+        // eslint-disable-next-line no-underscore-dangle
+        socketReady: connector?._transport?._socket?.readyState === 1,
+      }}
+    >
+      {children}
+    </context.Provider>
+  );
+};
+
+export default Provider;
```
gre, line 280: I have the feeling that this WalletConnect/Provider should be moved to live-common.

except some specific part it's not specific to mobile. we could do something similar to what we did for the "BridgeSync" in future if we generalize with desktop. (e.g. https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/bridge/BridgeSyncContext.js which connect the specific implementation of mobile to use the generic's provider implemented in live-common)

src/screens/WalletConnect/Scan.js
```diff
@@ -0,0 +1,48 @@
...
+type RouteParams = {
+  accountId: String,
+};
+
+type State = {};
+
+class ScanWalletConnect extends PureComponent<Props, State> {
+  componentDidMount() {
+    if (Config.MOCK_SCAN_WALLETCONNECT) {
+      setTimeout(async () => {
```
gre, line 23: this timeout should be cleared on Unmount in case user abort before.

gre (changes requested):

gre:
(btw it's all mostly suggestion and nothing that really impacts or stops the QA, all is good to me)

src/rootnavigation.js
```diff
@@ -0,0 +1,7 @@
+import * as React from "react";
+
+export const navigationRef = React.createRef();
```
gre, line 3: interesting idea :notbad:

src/screens/SignMessage/03-ConnectDevice.js
machard: yes I think so, I imagine it will be reused by desktop

src/screens/WalletConnect/Provider.js
machard: will change

src/screens/WalletConnect/Provider.js
machard: I will update, do I use this play store link https://play-lh.googleusercontent.com/RVKjd96rcTjiAnr45Gy6Nj2kCJ4_opdU2mrop7KftfyRhPWJf5ukvUR_Gi9AtOA920I or should I use something else ? @gre

src/screens/WalletConnect/Provider.js
machard: will change

src/screens/WalletConnect/Provider.js
machard: Yes I think we can move it to ledger live common

src/screens/WalletConnect/Scan.js
machard: will change

machard:
@gre Should we put Provider and the createAction in ledger-live-common now OR wait that this PR in the current state pass QA and do the move to live-common when adding the feature to ledger-live-desktop ?

gre:
It could be a quick win to do it now I think because that prevent you to have later to resend a PR in mobile, but it's not a blocker for the QA.

src/screens/WalletConnect/Provider.js
machard: some things need to be put out though, navigation logic and account selector.

Arnaud97234 (changes requested):
After a successful connection to wallet connect, while on WalletConnect Example screen:
- Kill LL then restart it
_Result: User land on dashboard. Still connected on desktop side_
- Click Disconnect on desktop end
- Start WalletConnect flow again on mobile end
_Result: LL crash with the following:_
IMG_1071

Arnaud97234 (approved):

gre (dismissed):
i'll solve the yarn.lock conflict and merge.

cryptoida:
Hi,
Where can i find  ledger-live-mobile compiled apk,  with WalletConnect implementation ?
If it is not done yet, can you please kindly compile ledger-live-mobile source with WalletConnect for android?

Thanks.

gre:
**Please beware these are unstable builds and NOT A RELEASE but our CI is currently producing builds on all commits.**

https://github.com/LedgerHQ/ledger-live-mobile/actions?query=workflow%3A%22bundle+app%22

mirshko:
Does this PR also allow users of the legacy Ledger (not the nano) to use WalletConnect with the desktop app? 

If all users of Ledger products can now use the WalletConnect integration we can remove all our code for the direct ledger integration from our site.

Let me know!

PR #1453, opened 2020-11-16
Move resources to dedicated screen for small screen comp

juan-cortes:
Due to the bug found by @nabil-brn where a user could be locked in the help button bottom modal if the display was too small, we moved the content of that modal to a dedicated screen. Title to be checked with product cc @dasilvarosa in case the one we came up with is not good enough

image

### Type

UI Polish/Bug fix

### Context

Slack https://ledger.slack.com/archives/GBX53P53Q/p1605519371031000

Make sure the screen works, the links link, and the back action too.

### DIFF
--- a/src/components/BottomModalChoice.js
+++ b/src/components/BottomModalChoice.js
@@ -17,7 +17,7 @@ const hitSlop = {
 };
 
 export default class BottomModalChoice extends PureComponent<{
-  onPress: ?() => void,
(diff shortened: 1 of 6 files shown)


gre (approved):

nabil-brn (approved):
✓ QA-ed

---
6c7a8cea2 (HEAD, juan/smallscreens) Wording changes

PR #1465, opened 2020-11-18
LL-3878 fix balance in first step of send flow

machard:
https://ledgerhq.atlassian.net/browse/LL-3878

### DIFF
--- a/src/components/AccountCard.js
+++ b/src/components/AccountCard.js
@@ -25,6 +25,13 @@ class AccountCard extends PureComponent<Props> {
     const { onPress, account, style, disabled } = this.props;
     const currency = getAccountCurrency(account);
     const unit = getAccountUnit(account);
+
+    const availableOnCompound =
+      account.type === "TokenAccount" && !!account.compoundBalance;
+    const balance = availableOnCompound
+      ? account.spendableBalance
+      : account.balance;
+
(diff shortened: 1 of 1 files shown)


src/components/AccountCard.js
```diff
@@ -25,6 +25,13 @@ class AccountCard extends PureComponent<Props> {
     const { onPress, account, style, disabled } = this.props;
     const currency = getAccountCurrency(account);
     const unit = getAccountUnit(account);
+
+    const availableOnCompound =
+      account.type === "TokenAccount" && !!account.compoundBalance;
+    const balance = availableOnCompound
+      ? account.spendableBalance
```
gre, line 32: mmh I'm not sure about the logic here, why are we not doing `account.spendableBalance` for all? I thought it was equal to balance in non compound case already.

machard: Ha ok. I was not sure if spendableBalance was available all the time

gre (approved):

nabil-brn (approved):
Reproduced on an old build and fixed on this one, the balance is well computed.

---
iOS (iPhone 11 Simulator)
d81f5a7fb (HEAD -> pr/machard/1465, machard/LL3878) spendableBalance is always set

PR #1466, opened 2020-11-18
LL-3857 change wording and condition of a lending banner

machard:
https://ledgerhq.atlassian.net/browse/LL-3857

### DIFF
--- a/src/screens/Account/LendingBanners.js
+++ b/src/screens/Account/LendingBanners.js
@@ -33,9 +33,9 @@ export default function LendingBanners({ account, parentAccount }: Props) {
   const compoundCapabilities = availableOnCompound
     ? getAccountCapabilities(account)
     : {};
-  const compoundSummary =
-    availableOnCompound &&
-    makeCompoundSummaryForAccount(account, parentAccount);
+  const compoundSummary = availableOnCompound
+    ? makeCompoundSummaryForAccount(account, parentAccount)
+    : null;
 
   let lendingInfoBanner = null;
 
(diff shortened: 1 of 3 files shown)


gre (approved):

Arnaud97234 (changes requested):
We should see the notEnough Disclaimer for account in the following configuration:
- Enable specific amount (Not max)
- Supplied: Amount supplied = Amount enabled
- Account available balance > Amount enabled

machard:
I don't understand, Supplied: Amount supplied < Amount enabled is not possible to pass the condition because of `getSupplyMax(account).eq(0)`

gre (approved):

src/screens/Account/LendingBanners.js
```diff
@@ -28,6 +33,9 @@ export default function LendingBanners({ account }: Props) {
   const compoundCapabilities = availableOnCompound
     ? getAccountCapabilities(account)
     : {};
+  const compoundSummary =
+    availableOnCompound &&
+    makeCompoundSummaryForAccount(account, parentAccount);
```
gre, line 38: i'm not sure if it passes Flowtype, but it is better for flow to do `availableOnCompound  ? make..() : null` otherwise it will tend to infer things like `CompoundSummary | boolean`

machard: looks like it's passing. However I get the point. I think it passes because compoundSummary is not typed ? I'll update

gre (approved):

PR #1470, opened 2020-11-18
AnalyticsConsole, debugging tool for tracked events

juan-cortes:
> Since this includes the same bumps and updates from https://github.com/LedgerHQ/ledger-live-mobile/pull/1468 I'm HODL'ing it until we can rebase it to not include all these changes.

This pr backports the analytics console concept from LLD to LLM, allowing us to view as an overlay on the screen the events that are getting tracked via analytics. This is specially useful for product to find the name of an event to polish metrics and to identify missing or incorrect events.
Group

It has already helped identify the fact that **we were missing all the events from the first launch of the app**. The component in charge of this was not getting the update of the `hasCompletedOnboarding` flag having changed and therefore we remained with uninitialised analytics for the duration of the first launch. Subsequent launched would work as expected, but we lost the metrics of all the first launches. This should be addressed here too.

### Type

Feature

### Context

n/a

- During onboarding, clicking on the Powered by ledger footer multiple times should take you to the debug screen.
- Enable the analytics console in the mock & test screen
- Complete the onboarding and take a look at events tracked.

Note: This behaves strangely with live reload so for best reliability don't edit the code while you are debugging the events. Relaunch the app completely if possible.

### DIFF
--- a/src/analytics/HookAnalytics.js
+++ b/src/analytics/HookAnalytics.js
@@ -1,35 +1,21 @@
-import { Component } from "react";
-import { connect } from "react-redux";
+import { useEffect, useState, useCallback } from "react";
+import { useSelector } from "react-redux";
 import { hasCompletedOnboardingSelector } from "../reducers/settings";
 import { start } from "./segment";
 
-let isAnalyticsStarted = false;
+const HookAnalytics = ({ store }: { store: * }) => {
+  const hasCompletedOnboarding = useSelector(hasCompletedOnboardingSelector);
+  const [analyticsStarted, setAnalyticsStarted] = useState(false);
 
-class HookAnalytics extends Component<{
-  store: *,
-}> {
-  componentDidMount() {
-    this.sync();
-  }
+  const sync = useCallback(() => {
+    if (analyticsStarted || !hasCompletedOnboarding) return;
+    setAnalyticsStarted(true);
+    start(store);
+  }, [analyticsStarted, hasCompletedOnboarding, store]);
 
-  componentDidUpdate() {
-    this.sync();
-  }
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -52,17 +52,17 @@
...
+    "@ledgerhq/hw-transport": "5.30.0",
     "@ledgerhq/hw-transport-http": "5.29.0",
-    "@ledgerhq/live-common": "16.0.0-countervalues.1",
-    "@ledgerhq/logs": "5.29.0",
-    "@ledgerhq/react-native-hid": "5.29.0",
-    "@ledgerhq/react-native-hw-transport-ble": "5.29.0",
+    "@ledgerhq/live-common": "17.0.0",
+    "@ledgerhq/logs": "5.30.0",
+    "@ledgerhq/react-native-hid": "5.30.0",
+    "@ledgerhq/react-native-hw-transport-ble": "5.30.0",
```
gre, line 65: (HODL this PR and wait for live-common to get merged because live-common 17.0.0 have bug on xrp)

gre (approved):

juan-cortes:
Rebased, updated, should be ok to QA at this point to me @gre

gre (approved):

nabil-brn (approved):
We've now access to the hidden settings at the start of the app and we're able to see analytic events.

Also I've noticed that if we add a debug http device before the onboarding, we could get stuck in the navigation but anyway it's not the scope of this ticket and a user is not supposed to use the hidden settings.

---
iOS & Android
7d17ec8cf (grafted, HEAD -> pr/juan-cortes/1470, juan/analytics) AnalyticsConsole, debugging tool for tracked events

PR #1471, opened 2020-11-19
fix name of app

valpinkman:
So now we can build a multitude of ios app bundle so it's possible to have various version of Ledger Live installed


gre (dismissed):
feel free to merge this whenever.

gre (approved):

PR #1473, opened 2020-11-19
Remove segment modification for resetting IP address

JunichiSugiura:


gre:
(context: the forked code we had was not helping anything, ongoing internal discussion)

src/analytics/segment.js
```diff
@@ -41,10 +41,6 @@ const extraProperties = store => {
   };
 };
 
-const context = {
-  ip: "0.0.0.0",
-};
```
gre, line 46: please rollback the change on JS side. it's actually a bug in the current segment lib but we must keep the present code.

gre (changes requested):
see comment

gre (approved):

PR #1474, opened 2020-11-23
LL-3877 add buy/receive info drawer on lending dashboard

machard:
Add an info drawer when a token account is non existant in the lending dashboard

### Type

Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-3877

You need to test with an eth account without having token accounts. and test after with one eth account which has token accounts.

### DIFF
--- a/src/screens/Lending/Dashboard/InfoModalBottom.js
+++ b/src/screens/Lending/Dashboard/InfoModalBottom.js
@@ -11,7 +11,7 @@ import TrackScreen from "../../../analytics/TrackScreen";
 
 type button = {
   title: string,
-  onPress: function,
+  onPress: Function,
 };
 
 type Props = {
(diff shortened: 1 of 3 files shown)


gre (changes requested):
should explicit the function type (e.g. `() => void`)

src/screens/Lending/Dashboard/InfoModalBottom.js
```diff
@@ -11,7 +11,7 @@ import TrackScreen from "../../../analytics/TrackScreen";
 
 type button = {
   title: string,
-  onPress: function,
+  onPress: Function,
```
gre, line 14: preferably we should specify exactly the signature btw, because `Function` is like `any` function. but ok for now.

gre (approved):

Arnaud97234 (changes requested):
### Bug found

Application crashes when selecting the ETH account in the Buy flow.
It happen when accessing buy flow from the Lending flow. Works ok when going to Buy from the bottom bar.

appcrash_BuyFromLend

gre (approved):

PR #1485, opened 2020-11-30
added devices modelIds to event properties

machard:
https://ledgerhq.atlassian.net/browse/LL-3901


src/analytics/segment.js
```diff
@@ -39,6 +40,7 @@ const extraProperties = store => {
     platformVersion: Platform.Version,
     sessionId,
     devicesCount: devices.length,
+    devicesVersions,
```
gre, line 43: segment DO NOT support arrays of data. this would badly break and never track any analytics if i recall correctly.

in any case, we need to follow the same impleemntation as the one we will do on LLD where i had some proposal of changes (see other PR on LLD)

gre (changes requested):
- not enough in order to be consistent with

src/analytics/segment.js
machard: ha ok, i looked quickly and I saw "The Schema will show you how recently we’ve seen your event properties and their type (ex: array, boolean, data, number, text)"

gre: yeah i think in practice it is not properly integrated (maybe due to the "destinations" we use)

src/analytics/segment.js
```diff
@@ -27,7 +27,13 @@ const extraProperties = store => {
   const state: State = store.getState();
   const { localeIdentifier, preferredLanguages } = Locale.constants();
   const devices = knownDevicesSelector(state);
-  const devicesVersions = devices.map(device => device.modelId);
+  const lastDevice = devices[-1];
```
gre, line 30: this is always undefined 🤔

gre (dismissed):
please check feedback because i think it doesn't work

machard:
@gre can it be because of https://github.com/LedgerHQ/ledger-live-mobile/pull/1506 ?

cthepot-ledger:
@machard not passing the CI

machard:
linting passed, build not, im not sure that the builds are working rn in PR.
I merged develop, lets see

juan-cortes:
> @gre can it be because of #1506 ?
> (I only have the simulator with the ws device so I can not really check the pairing and see if the knowDevices list is currently populating or not)

If it's still working you should be able to pair mock devices if you have the mock flag enabled on LLM. It should list you a couple of devices such as "David's nano X" or similar, you should be able to debug this in the future if needed.

src/analytics/segment.js
```diff
@@ -27,6 +27,13 @@ const extraProperties = store => {
   const state: State = store.getState();
   const { localeIdentifier, preferredLanguages } = Locale.constants();
   const devices = knownDevicesSelector(state);
+  const lastDevice = devices[devices.length - 1];
+  const deviceInfo = lastDevice
+    ? {
+        deviceVersion: lastDevice.deviceInfo?.version,
```
gre, line 33: consistently with desktop, shouldn't this also have modelId? https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/analytics/segment.js#L46

machard: I assumed it's always nanoX on mobile.
I have added

gre (dismissed):

gre:

PR #1489, opened 2020-12-01
add a utxo warning when utxo > 50

machard:
https://ledgerhq.atlassian.net/browse/LL-2847

i've seen the feedback on lld, i will update

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -1598,6 +1598,11 @@
     "loading": "Searching devices..."
   },
   "send": {
+    "tooMuchUTXOBottomModal": {
+      "cta": "Continue",
+      "description": "This transaction may take long to verify and sign because the account has a significant number of coins.",
+      "title": "UTXO message"
+    },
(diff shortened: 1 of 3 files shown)


Arnaud97234:
@gre @machard In my opinion it's a go. The only comments I have:
- Transactions can take a couple of seconds to be displayed on the device after "Loading" reach 100%
- Is there any chance that the device goes to sleep before the tx is displayed on the device? (I can dig more on that one if it worth it)

gre:
@Arnaud97234 these are good point.

for (1) yeah, i think there are improvements into that, but it's a tricky topic. it's in ledgerjs logic. a bit unrelated to that PR but it's useful feedback still. a bit different but see also https://ledgerhq.atlassian.net/browse/LL-3840 , it indicates we can do a bit better in term of % feedback.

for (2) yeah, probably to be discussed with Firmware team as part of the "please wait before making device go sleep mode" kind of situation (install apps, scan accounts,.. are also 2 cases you may experience this problem already).

gre (approved):

PR #1499, opened 2020-12-07
LL-4008 WC icon

machard:
https://ledgerhq.atlassian.net/browse/LL-4008

(I could also put the screen in loading state instead of this orange indicator.)

to test:
just do a walletconnect and it will appear briefly.


gre (approved):

gre:
to faciliate testing, i've included in #1505

PR #1503, opened 2020-12-10
Fixes XRP send

gre:
we recently did a regression due to the fact we didn't upgrade `buffer` library but we upgraded `ripple-lib` which started using a more recent call that was not yet defined by buffer. ( this is the same issue as described here https://github.com/ripple/ripple-lib/issues/1319 )

included work https://github.com/LedgerHQ/ledger-live-common/compare/hotfixes/17.2.x

### Type

bug fix

### Context

users reporting the issue on twitter

- send XRP
- also generally check other general families are fine (eth, bitcoin) as `buffer` is used everywhere
- includes the workaround for countervalues graph missing in bad case of missing old bitcoin ops


juan-cortes (approved):

gre:
@Arnaud97234 this has been fixed on Android.

Arnaud97234 (commented):
I reproduced the `WebSocket` in an edge case on Android (1 tx out of 10 attempts): I have the same accounts on 2 platform

Arnaud97234 (approved):

gre:
ok that can be a first shippable patch for 2.18.1 and we hopefully work on remaining glitch for 2.19 🤞

PR #1506, opened 2020-12-13
Fix bug on autoSelectOnAdd manager pairing

juan-cortes:
image

A recently discovered bug was preventing the `autoSelectOnAdd` functionality from working, presumably in the whole app. The pairing flow will eventually return just the device id instead of the expected `Device` and in doing so would fail any subsequent device action linked to the particular screen. This meant that the manager flow, ~the onboarding flow~, and the swap flows would fail to connect to a newly paired device. 

This isn't a dead end, since the user can close the bottom modal, and select the newly paired device. Which would then connect without issues. But it does align with many of the connectivity issues we've been reporting/experiencing lately. Basically, if you pair a new device on the current develop it will give you an error when trying to connect, until you dismiss the modal and connect again.

This pr fixes that.

In the same process, I've identified a few wording inconsistencies that were introduced in the LL-3860 task, were we had a gazillion wording changes, some of which have been identified as incorrect, at least to me.

### Type

Bug fixes

### Context

Slack

First make sure you can reproduce the issue on develop so that you see what I'm talking about.
- Unpair a nano x
- Go to the manager/swap flow
- Add a new nano x
- On the pair success screen (where you have an option to rename it) click on continue
- The bottom modal will open automatically upon returning to the flow screen
- The connection will fail.

Then on this branch try the same process
- Unpair a nano x
- Go to the manager/swap flow
- Add a new nano x
- On the pair success screen (where you have an option to rename it) click on continue
- The bottom modal will open automatically upon returning to the flow screen
- The connection will work.

### DIFF
--- a/src/components/SelectDevice/index.js
+++ b/src/components/SelectDevice/index.js
@@ -96,7 +96,7 @@ export default function SelectDevice({
         return devices;
       });
     });
-    return () => subscription.unsubscribe;
+    return () => subscription.unsubscribe();
   }, [knownDevices, filter]);
 
   return (
--- a/src/components/DeviceItemSummary.js
+++ b/src/components/DeviceItemSummary.js
@@ -15,7 +15,6 @@ import Alert from "../icons/Alert";
 
 type Props = {
   deviceId: string,
-  name: string,
   genuine: boolean,
   onEdit: () => void,
 };
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -1944,7 +1944,7 @@
     },
     "GenuineCheck": {
       "title": "Device authentication check",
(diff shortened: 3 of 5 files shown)


src/components/SelectDevice/index.js
```diff
@@ -96,7 +96,7 @@ export default function SelectDevice({
         return devices;
       });
     });
-    return () => subscription.unsubscribe;
+    return () => subscription.unsubscribe();
```
JunichiSugiura, line 99: Thanks 🙇‍♂️

juan-cortes:
@Arnaud97234

Arnaud97234 (approved):

PR #1518, opened 2020-12-18
pass correct remote when pushing android tags

valpinkman:
Just a small fix to make sure the remote is `upstream` when pushing git tags in android workflow


juan-cortes (approved):
lgtm

nabil-brn (approved):
No QA needed, correct me if i'm wrong.

PR #1528, opened 2020-12-28
LL4225 - remove link to account when in walletconnect mode

machard:
https://ledgerhq.atlassian.net/browse/LL-4225

### DIFF
--- a/src/screens/OperationDetails/Content.js
+++ b/src/screens/OperationDetails/Content.js
@@ -1,5 +1,5 @@
 /* @flow */
-import React, { useCallback, useState } from "react";
+import React, { useCallback, useState, useContext } from "react";
 import { View, StyleSheet, Linking } from "react-native";
 import uniq from "lodash/uniq";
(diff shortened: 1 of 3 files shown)


gre (dismissed):
I find the implementation a bit weird (opération détails don't strictly depends on walletconnect)

I would prefer we add in the navigation object a flag like "disableAllNavigations"

machard:
when launching the send fund flow, or when going to the operation details page ?

(I usually go with the easy way until it makes my life complicated ^^. I would say that I consider creating a new concept when the easy way starts to be annoying)

gre:
I think you simply can add it in the `navigate("OperationDetail", { flag here })` (pseudo code)
i think it's better because that makes the screens more uncoupled (and for instance, we could test the screen individually without it being implicitly depending on WC)

gre (approved):

Arnaud97234 (approved):

gre:

PR #1536, opened 2021-01-05
LL-4260 Setup a new device, remove tab bar

juan-cortes:
This bug affected both accessing the feature from the read only mode after the onboarding (is it still a thing?) and access the feature from the settings after the onboarding. The reason for this was that we had defined the inner onboarding navigator as a screen for settings/manager navigators instead of at the top level above tab bar navigation. Rearranging this allows us to navigate to the child screen with whatever params without compromising the layout (without showing the tab bar, because we are not inside a tab bar navigator anymore)

### Type

Bug fix.

### Context

https://ledgerhq.atlassian.net/browse/LL-4260

- Normal onboarding should work as expected and **not** have tab bar.
- Accessing the onboarding from a read only ios device **after** successful onboarding should **not** have tab bar.
- Accessing the onboarding from the settings/help screen should .... you know the drill!
- Going **back** to the previous screen should also work, and not some other screen.


gre (approved):

gre:
conflicts. feel free to ask @LFBarreto  (dark mode related i think)

gre:

PR #1537, opened 2021-01-05
LL-1944 POC of log view/export for LLM

juan-cortes:
I'm not even sure why this was assigned to me in the first place but to get the ball rolling this is at least better than nothing. It providers a screen for us to see the logs (for users, I guess, if we put it out of the debug menu) and a way to export the logs. Currently all very rough with just copying the logs as a huge string but I guess we could generate a file and share that instead.

I think that with some modifications to the way we dispatch the logs [here](https://github.com/LedgerHQ/ledgerjs/blob/fd8137d029835f3e8966d6051fb0e957eab194a5/packages/logs/src/index.js#L24-L29) we could retrieve logs from before accessing the screen, think a buffer of past N logs, or a replay rxjs thingy.

image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1944


gre (approved):

gre:

please do another PR for the mentioned fix

PR #1554, opened 2021-01-20
LL-4344 Drop unused literals

juan-cortes:
Brings the hacky script for finding unused literals in the app from desktop into mobile. The working is straightforward enough, you run `python3 scripts/i18next-unused.py` and it will parse all the literals, and look for the recursive key that's built. For dynamic keys such as `im.a.${dynamic}.key` I currently added a list of regular expressions so that the script is somewhat resillient to dynamic keys. If all fails, there's an exclude array, to exclude some keys from the list.

As with the other pr, heavy testing would be appreciated, and if devs can also take a look at the dropped keys to see if something stands out, it would minimise the possibility of regressions.

### DIFF
--- a/scripts/i18next-unused.py
+++ b/scripts/i18next-unused.py
@@ -1,0 +1,144 @@
+#! /usr/local/bin/python3
+
+import os
+import json
+from sys import argv
+import sys
+import re
+
+folder1 = './src'
+locales = './src/locales/en/common.json'
+files = []
+rawkeys = {}
+keys = {}
+excludedKeys = [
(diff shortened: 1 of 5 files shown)


gre (approved):

juan-cortes:
At this point, and after more than two months of delay, after having at least two false positives on LLD. 
I don't feel confident the dropped literals are all really dropped. I think it's better we leave it as is (perhaps keep the script) until we have more coverage of the app with tests to prevent issues.

cc @gre  to me it's a close for now

PR #1558, opened 2021-01-26
LL-4432 Terms localized

LFBarreto:
Mainly upload translated term files to :octocat: and add logic to show the correct one according to the language selected

### Type

Feature
### Context

LL-4432

Terms and condition once translation are enabled


cthepot-ledger:
@LFBarreto Can you check the CI?

gre (changes requested):
please split this PR in two parts:
- first we must land the TERMS on develop & then master.
- second we can actually test the PR itself, otherwise QA have no mean to test this.

gre:
the flag "important" was meant to say "it's to be merged now". right now, it's not ready.

gre (approved):

PR #1559, opened 2021-01-26
LL-4429 remove sub accounts for ethereum classic

machard:
https://ledgerhq.atlassian.net/browse/LL-4429

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -232,6 +232,7 @@ GEM
 PLATFORMS
   ruby
   universal-darwin-20
+  x86_64-darwin-19
 
 DEPENDENCIES
   cocoapods (= 1.8.4)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -232,7 +232,6 @@ GEM
 PLATFORMS
   ruby
   universal-darwin-20
-  x86_64-darwin-19
 
 DEPENDENCIES
   cocoapods (= 1.8.4)
--- a/src/screens/Account/SubAccountsList.js
+++ b/src/screens/Account/SubAccountsList.js
@@ -303,9 +303,7 @@ export default function SubAccountsList({
 
   if (
     !isToken &&
-    subAccounts.length === 0 &&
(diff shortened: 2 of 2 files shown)


gre (changes requested):
mmh, I think it would be better if we could make live-common not  having any value for `account.subAccounts` , so I think it's a fix to do in live-common's families/ethereum/* and specifically on the "synchronisation.js" to only have that subAccounts array for ethereum

machard:
@gre the only reason i had to add the "ethereum" check is **because** of the "tezos" check.

==> this is enough to avoid displaying token accounts on ethereum classic without touching llc ==>
`if (
    !isToken &&	    !isToken &&
    subAccounts.length === 0 &&	    subAccounts.length === 0)`
  
  I don't know why there is a "tezos" check. I think it's been added by @juan-cortes if I remember my histories research of last week?

gre:
yeah, because essentually we would never render that component for Tezos that have subAccounts undefined as well as ethereum_classic

see https://github.com/LedgerHQ/ledger-live-mobile/blob/1dd9696b7372205fa831c1eda6353af4f62c7f17/src/screens/Account/ListHeaderComponent.js#L200

so essentially we need to fix that in live-common:

- find a way to put subAccounts: undefined if subAccounts.length===0 for tezos (imo you can do a postBuildAccount like we had to customize for bitcoin family https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/bitcoin/libcore-postBuildAccount.js )
- find a way to also not set subAccounts for all ethereum family except ethereum itself (Ethereum Ropsten also don't have tokens in our context)

machard:
@gre 

Isn't it enough ? 

https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/screens/account/TokensList.js#L102-L109

this is basically the same code as mobile without the "tezos" and "ethereum" check. If it works on mobile it should work on mobile yes ?

machard:
Ok so now it's the same code as in desktop so the behavior should be the same. the tezos "check" on mobile must have been a left over.

gre (approved):

PR #1563, opened 2021-01-26
LL-4449 Fix chart y-axis rendering

juan-cortes:
Improve the chart so that ranges with small variations don't start with zero, also allowing to have a small padding from the lowest value of Y. In the process I found a small glitch where the hovered amount would make the screen jump ever so slightly, so that bug has also been addressed. It's most noticeable in the empty account and the third screenshot pair where we'd originally start by zero and see the changes as very minor.

This alongside https://github.com/LedgerHQ/ledger-live-desktop/pull/3516 solves https://github.com/LedgerHQ/ledger-live-desktop/issues/3507

## Before - After graphs
image

## Hover bug
image

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-4449


gre (approved):

nabil-brn (approved):
✓ QA-ed

---
9ecfa5584 (HEAD -> LL-4449, juan/LL-4449) Merge branch 'develop' into LL-4449

PR #1570, opened 2021-01-28
Live common 18.3.x

gre:
https://github.com/LedgerHQ/ledger-live-common/releases/tag/v18.3.1

closes #1562


Arnaud97234 (changes requested):
We still having a bug on the error message when swap amount is too low:
- DGB magnitude is 8 so we should do x100000000 on the value currently displayed in the error message.
- Also When selecting BTC or ETH the error just say “amount should be at least 0 ETH”

IMG_1132

IMG_1133

Arnaud97234 (approved):

PR #1574, opened 2021-02-02
LL-4532 Remove backup pack banner

juan-cortes:
### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-4532

- Open the app
- Make sure the banner is not present

### DIFF
--- a/src/components/Carousel/index.js
+++ b/src/components/Carousel/index.js
@@ -14,7 +14,7 @@ import Card from "../Card";
 import LText from "../LText";
 import Button from "../Button";
 import BuyCrypto from "./slides/BuyCrypto";
-import BackupPack from "./slides/BackupPack";
+// import BackupPack from "./slides/BackupPack";
 // import BlackFriday from "./slides/BlackFriday";
 // import Sell from "./slides/Sell";
 // import Vote from "./slides/Vote";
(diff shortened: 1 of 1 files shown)


LFBarreto (approved):

gre (changes requested):
@juan-cortes is there a reason why we are not removing the code here? If these are not coming again in future, there is no reason why we are commenting. we should just delete all of these.

In this particular case the banner will come back once there's stock if I understood correctly what @cthepot-ledger said, so it made sense to just comment it out. There seem to be other banners commented out, I'm all up for deleting them, and the associated files, and recovering them from git if needed in the future

gre (approved):
ok

juan-cortes:
> @juan-cortes is there a reason why we are not removing the code here? If these are not coming again in future, there is no reason why we are commenting. we should just delete all of these.



In this particular case the banner will come back once there's stock if I understood correctly what @cthepot-ledger said, so it made sense to just comment it out. There seem to be other banners commented out, I'm all up for deleting them, and the associated files, and recovering them from git if needed in the future

nabil-brn (approved):
✓ QA-ed

---
45148026f (HEAD -> LL-4532, juan/LL-4532) LL-4532 Remove backup pack banner

PR #1581, opened 2021-02-09
Coin 1146 network fees validate device

henri-ly:
### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/COIN-1146

Every modal, by default we are using the same as before, but it will happened in that some case we will need to add some extra info box


rjarasson-ledger (approved):

gre:
shipped through mashup pr

PR #1582, opened 2021-02-09
SENTRY - fix sentry crash

machard:
https://sentry.io/organizations/ledger/issues/2207923304/?project=273101&query=is%3Aunresolved&statsPeriod=14d

the component was called without color here https://github.com/LedgerHQ/ledger-live-mobile/blob/1dd9696b7372205fa831c1eda6353af4f62c7f17/src/components/RootNavigator/BaseNavigator.js#L211

### DIFF
--- a/src/components/RootNavigator/BaseOnboardingNavigator.js
+++ b/src/components/RootNavigator/BaseOnboardingNavigator.js
@@ -22,7 +22,8 @@ const hitSlop = {
   top: 10,
 };
 
-export const ErrorHeaderInfo = ({ route, navigation, colors }: *) => {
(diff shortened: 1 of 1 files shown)


gre:
@machard do you have a way to test it for QA to confirm it's fixed?

LFBarreto:
To reproduce you try add a new device via bluetooth once the device is found and in the list lock it back and proceed to select it in order to get an error during the pairing that would trigger the info icon on top of the page. (which was not showing correctly and was the main bug here)

gre:
shipped through mashup pr

PR #1588, opened 2021-02-10
LL-4396 (Onboarding): fix button tap styling

LFBarreto:
(Onboarding): switch from Pressable to TouchableOpacity for built in on press opacity styling since Pressable does not make sense here.
### Type

UI Polish
### Context

LL-4396

Onboarding buttons should now react in opacity when pressed


juan-cortes (approved):
lgtm

PR #1604, opened 2021-02-24
LL-4520 (Modals): block opening two modals at same time

LFBarreto:
(Modals): block opening two modals at same time
### Type

Bug fix
### Context

LL-4520

Tap quick on two modal buttons per instance in the accounts page header buttons or the transfer button at the same time.
It should no longer open both modals.

### DIFF
--- a/src/components/BottomModal.js
+++ b/src/components/BottomModal.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { createRef, useEffect, useState } from "react";
+import React, { useEffect, useState } from "react";
 import { View, StyleSheet, Platform } from "react-native";
 import ReactNativeModal from "react-native-modal";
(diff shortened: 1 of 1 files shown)


src/components/BottomModal.js
```diff
@@ -11,6 +11,8 @@ import ButtonUseTouchable from "../context/ButtonUseTouchable";
 import getWindowDimensions from "../logic/getWindowDimensions";
 import DebugRejectSwitch from "./DebugRejectSwitch";
 
+const isModalOpenedref = createRef();
```
gre, line 14: there is no need for this to be a ref, it can just be a `let isModalOpened` global variable.

valpinkman (approved):

Arnaud97234 (approved):

PR #1613, opened 2021-03-02
LL-4537 input clear button for dark/dusk themes

juan-cortes:
Stops using the native clear button for iOS devices since we can't specify the theme that we are using. Mirroring the implementation from the Android component who implement our own clear button that gets the right color from the theme.

https://user-images.githubusercontent.com/4631227/109646254-40d56f00-7b58-11eb-8a92-8d3cb974ae3b.mov

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-4537

Although not used in _many_ parts, inputs are present throughout the app and we need to make sure that this has not introduced any regressions in them. So for the sake of clarity, make sure the inputs still work in at least:
- Send flow
- Add account flow
- Manager search
- Manual fees input


LFBarreto (approved):
LGTM could you check if this implementation can also work for android if so maybe we can unify both TextInput into one :shrug:

juan-cortes:
> LGTM could you check if this implementation can also work for android if so maybe we can unify both TextInput into one 🤷

Sadly the logic still differs for the keyboard type, I mean, it could be done and it might not be that hard but certainly not as it stands right now. Passing the keyboard type needed for android on the input fields, crashes on ios because it's not a valid type. So we'd end up with platform specific stuff in the component anyway. 
Could fit a new ticket for cleaner code though

Arnaud97234 (changes requested):
### Bugs found
1) Cross should be in the input field. it shouldn't  appear hover the cancel CTA.

2) Keyboard should be displayed when tapping in the input field

Actual:
image

Expected:
image

nabil-brn (changes requested):
Clear button require 2 tap (1st to remove the keyboard, 2nd to clear the field) on manager search bar

https://user-images.githubusercontent.com/16896559/110735033-48b4a380-8229-11eb-8bfb-9262d1ba5dc5.mov

Middle of the clear button triggers cancel button on manager search bar

https://user-images.githubusercontent.com/16896559/110734953-1dca4f80-8229-11eb-9e17-8f040d9227a9.mov

I just notified that the clear button changed in this part of UI (hidden menu)
Capture d’écran 2021-03-11 à 05 04 32

---
767b6b03d (HEAD -> LL-4537, juan/LL-4537) Fixes for ios input ref

juan-cortes:
oh my -.-

juan-cortes:
I think it's best if we just revisit the whole component as @LFBarreto suggested some time ago and unify it between Android and iOS, this was supposed to be a quick win for dark themes and it clearly hasn't been. I've moved the issue back to todo and I would suggest whomever takes the task to disregard these changes.

PR #1617, opened 2021-03-03
LL-4768 Prevent sync if no network

LFBarreto:
(Network): prevent sync if no network - workarround for ios sync crash
### Type

Bug Fix
### Context

LL-4768

Workarround the ios sync crash bug on libcore related coins if no network


Arnaud97234 (changes requested):
It prevent only the crash on pull to refresh but when a sync is launched automatically the app crash, then keeps crashing when relaunching.
Also it happen during the add account flow, when the scan from the device starts the application crash.

Could we show an error on portfolio and accounts screen, and also totally block the sync when the Network is down ?

LFBarreto:
Closing this since the main issue wil be fixed on libcore

PR #1620, opened 2021-03-04
fix(AccountActions): wait BottomModal hide before navigate to avoid kb blur - LL-4696

rjarasson-ledger:
fixes https://ledgerhq.atlassian.net/browse/LL-4696

### Type

Bug Fix

### Context

AccountActions are displayed as a BottomModal list of choices, that redirects to different flows / screens.
Since BottomModal is animated, it is finally hidden after the screen is display.
For some reason, if the next screen uses a keyboard (autofocus), the BottomModal animation end leads to Keyboard being dismissed. This can have other unwanted bugs.

To avoid any conflict, this AccountActions components retains the next screen to go to, and waits for the Modal to be hidden before transitioning to next screen.

This leads to a slight delay, but avoids conflicts.

For Polkadot (and others having additional actions):
* tap on the more button
* tap on bond
* the bottom modal closes
* the bond flow starts

**I did not test on Android (only iOS).**

### Notes

This delay can be unwanted.
Other solution would be to add ref capabilities to CurrencyInput, then trigger the focus manually after a slight delay.
It's more complicated and specifically fixes the issue only for Polkadot - issue could reappear on other coins/flows.

I was unable to find why the keyboard is dismissed. `react-native-modal` uses a `InteractionManager.createInteractionHandle();` that could have this kind of side-effect.

### DIFF
--- a/src/screens/Account/AccountActions.js
+++ b/src/screens/Account/AccountActions.js
@@ -55,6 +55,7 @@ export default function AccountActions({ account, parentAccount }: Props) {
 
   const onNavigate = useCallback(
     (name: string, options?: NavOptions) => {
+      setNext();
       navigation.navigate(name, {
         ...options,
         params: {
@@ -86,7 +87,8 @@ export default function AccountActions({ account, parentAccount }: Props) {
 
   const goToNext = useCallback(() => {
     if (next) {
-      onNavigate(...next);
+      // workaround for bottom modal + text input autoFocus issue
+      setTimeout(() => onNavigate(...next), 0);
     }
   }, [onNavigate, next]);
 
--- a/src/screens/Account/AccountActions.js
+++ b/src/screens/Account/AccountActions.js
@@ -37,6 +37,7 @@ type NavOptions = {
 export default function AccountActions({ account, parentAccount }: Props) {
   const { colors } = useTheme();
   const [displayedActions, setDisplayedActions] = useState();
+  const [next, setNext] = useState();
   const navigation = useNavigation();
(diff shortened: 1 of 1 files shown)


rjarasson-ledger:
Seems to be not working on Android. Still work to do on this.

rjarasson-ledger:
Added a setTimeout (sorry ugly), it was still not working on android. Looks like this side-effect occurs just a frame after react-native-modal calls the onModalHide callback.

src/screens/Account/AccountActions.js
```diff
@@ -54,7 +55,7 @@ export default function AccountActions({ account, parentAccount }: Props) {
 
   const onNavigate = useCallback(
     (name: string, options?: NavOptions) => {
-      setDisplayedActions();
+      setNext();
```
gre, line 58: mmh i'm not sure to understand what this do. next is always undefined, a setNext(undefined) is a noop, react won't do anything because it's === 🤔

src/screens/Account/AccountActions.js
```diff
@@ -128,7 +137,8 @@ export default function AccountActions({ account, parentAccount }: Props) {
                     key={i}
                     onSelect={({ navigationParams, enableActions }) => {
                       if (navigationParams) {
-                        onNavigate(...navigationParams);
+                        setNext(navigationParams);
```
gre, line 140: oh it's set here, nevermind

gre (approved):

Arnaud97234 (approved):

PR #1641, opened 2021-03-24
[LL-4594] cosmos: display error on claim reward

henri-ly:
### Type

Bugfix

### Context

https://ledgerhq.atlassian.net/browse/LL-4594

Cosmos display error for claim reward


rjarasson-ledger (approved):

rjarasson-ledger (commented):
Shouldn't the color of warning be orange instead of "alert" then ?

henri-ly:
> Shouldn't the color of warning be orange instead of "alert" then ?

Probably, should i fixed it ?

Arnaud97234 (approved):
We should adapt the wording in order to make it more explicit for the user, for instance "Sorry, Insufficient funds to pay the fees" instead of just "Sorry, insufficient funds".
@cthepot-ledger

PR #1642, opened 2021-03-25
[LL-4851] feat: add Alert component + LightBulb & HandShield icons

rjarasson-ledger:
### Type

UI Polish

### Tickets
EPIC: https://ledgerhq.atlassian.net/browse/LL-4441
https://ledgerhq.atlassian.net/browse/LL-4851
FIX https://ledgerhq.atlassian.net/browse/LL-4566

### Context

The new Design System is being polished, and some existing components having same/similar styles should be refactored to use the design system components.

We start with the Alert component that will replace many existing components.
It inherits mainly from the `InfoBox` component.

Figma: https://www.figma.com/file/f3Q65y0n7sCnN9PsCMrhz6/Live-Library?node-id=302%3A42

### Specs

Alert is the new component for giving semantic feedback in the form of a colored Box, with a semantic icon on the left, an optional bold title, a text (single line only), and a right component, all vertically centered with margin inbetween.

Props:
- `title`: Optional title for the first line of alert (semi-bold text)
- `left`: replace the icon by anything (used to replace the icon if needed)
- `noIcon`: do not display the icon
- `children`: the alert content

We can display a "learn more" link at the end of the content (same line):
- `onLearnMore`: callback when click on learnMore
- `learnMoreUrl`: alternative to using callback (onLearnMore still take precedence)
- `learnMoreKey`: customize learnMore label with translation key (defaults to translation `common.learnMore`)
- `learnMoreIsInternal` - boolean: remove the external icon - defaults: false

It has a limited list of variants:
* `primary`: info circle icon - blue
* `secondary`: info circle icon - grey (NOTE: used as *default*)
* `success`: check circle icon - green
* `warning`: exclamation circle icon - orange
* `error`: cross circle icon - red
* `hint`: new LightBulb icon - blue
* `security`: red shield big icon - grey
* `help`: blue shield big icon - grey
* `danger`: red shield big icon - red text, paper background
* `update`: white exclamation triangle icon - white text, orange background (*NOTE: color in figma is not present in palette, so i used the existing orange*)

NOTE: contrary to Desktop Alerts, the following is different:
* no `small` variant (not relevant on mobile)
* no `right` (not enough space on mobile)
* no `learnMoreOnRight` variant, since no right...
* added a `vertical` variant

We could add a "bottom" prop to add something on the bottom, with maybe a `learnMoreOnBottom`, to replace the `right` feature on desktop.

* New Alert component for all styles:
* The DangerBox used in swap is replaced by a vertical "danger" alert:
  * before:  after: 
  * The bug ticket found in the process: https://ledgerhq.atlassian.net/browse/LL-5075 (the address is white on white on dark mode)
* Fix Polkadot UnsupportedWarning - THIS IS TEMPORARY since we have another ticket to add behaviour on this one 
* WarningBox removed and migrated to Alert type="warning"
* InfoBox removed and migrated to Alert type="primary"
  * Example for Tron (send max): 
* VerifyAddressDisclaimer removed and migrated to Alert type="help" or Alert type="danger": 
* Migrate AddAccountsTokenCurrencyDisclaimer disclaimer to Alert type="primary". 
 - *NOTE: because the "event" is not handled by the Alert component, we will loose tracability of the event `AddAccountsTokenDisclaimerLearnMore`.*
 - *NOTE2: I fixed the paddings too*
  - before: , after: 
* Experimental settings Disclaimer to vertical Alert type="security":
  - before:  after: 
* WarningBanner uses Alert type="update": 
* Removed the OngoingScams banner


rjarasson-ledger:
I added a `bottom` prop to allow adding anything in the bottom of the alert content.
I could be used to add a CTA specific to the alert.

rjarasson-ledger:
Rebased on develop just after Notifications Center merge:

* I did not use the new WarningOutline icon because it does not fit the figma neither. We really should realign icons (style, naming and size), e.g. for this warning triangle (TriangleWarning on LLD), we have on LLM: Warning, WarningOutline, AlertTriangle.
* I did not migrated the notifications alert-lookalike components because:
  - it's brand new and i haven't yet seen examples
  - it has an incident list that might not be possible to display with Alert as-is

rjarasson-ledger:
Also note on the "learn more link":

* This redesign does not provide a way to set an "event" with "eventProperties" for tracking.
* It does not user the "ExternalLink" component because a was not able to align and keep the right fontSize 

* Hence there is not opacity change on the learnMore link.

rjarasson-ledger:
Also, content cannot be multiline (yet ?).
Maybe if there is an existing way to split `\n` into multiple paragraph, this could be added easily.
I feel there will be a need for having multiple lines, but this would also be discutable with the learnMore placement...

rjarasson-ledger:
I realigned the separated Stash/Controller warning with the desktop version. It should enable adding a "Set controller" button to be added in the bottom.

Arnaud97234 (approved):

gre (approved):

PR #1654, opened 2021-04-01
LL-4626 - Add the possibility to retrieve latest log in mobile app

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LL-4626
Retrieve last N log in mobile app.
Currently, just for demo, I display the last N log when the button "Display latest logs" button is clicked in "Debug Logs" page.
But this logic can be integrated into any other place in the app

### DIFF
--- a/src/components/GenericErrorView.js
+++ b/src/components/GenericErrorView.js
@@ -2,10 +2,16 @@
 import React, { PureComponent } from "react";
 import { View, StyleSheet } from "react-native";
 
+import Share from "react-native-share";
+import { Trans } from "react-i18next";
 import LText from "./LText";
 import ErrorIcon from "./ErrorIcon";
 import TranslatedError from "./TranslatedError";
 import SupportLinkError from "./SupportLinkError";
+import logger from "../logger";
+import logReport from "../log-report";
+import Button from "./Button";
+import DownloadFileIcon from "../icons/DownloadFile";
 
 class GenericErrorView extends PureComponent<{
   error: Error,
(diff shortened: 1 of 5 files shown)


src/log-report.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable no-console */
+// @flow
+import { listen } from "@ledgerhq/logs";
+
+const logs = [];
+const logLimit = 10; // the number of latest log we want to conserve
```
gre, line 6: That value should be something like 1000
Similarly to desktop

src/screens/DebugLogs.js
```diff
@@ -38,6 +38,11 @@ export default function DebugLogs() {
     }
   };
 
+  const onDisplayLatestLogs = () => {
+    const logs = logReport.getLogs();
+    alert(logs.map(log => JSON.stringify(log)).join("\n"));
```
gre, line 43: I imagine next step is to use the Share API to let the user download it?

Example using this library https://github.com/react-native-share/react-native-share/issues/411#issuecomment-442220259

gre (changes requested):

src/screens/DebugLogs.js
hzheng-ledger: yes, also, I assume that the download log page (or download button) should appear when user encounters an error. So we need some implementation to catch the error in the app and put this feature in a correct place

juan-cortes: The DebugLogs screen I wrote already has the logic to export the logs, you just need to pass in the right data instead of the dummy one. As to the missing implementation, it's a matter of modifying the UI of the error screen/modal to show a button that triggers that same behaviour. Don't worry about catching more errors, that's out of the scope of the task

hzheng-ledger:
log
current screenshot

src/components/GenericErrorView.js
```diff
@@ -15,16 +21,40 @@ class GenericErrorView extends PureComponent<{
...
+        logger.critical(err);
+      }
+    }
+  };
   render() {
     const { error, outerError, withDescription, withIcon } = this.props;
     const titleError = outerError || error;
     const subtitleError = outerError ? error : null;
+    const hasExportLogButton = this.props.hasExportLogButton !== false; // hasExportLogButton is true by default
+
```
hzheng-ledger, line 57: Currently, I put hasExportLogButton = true by default. It means by default, an error message has this "save logs" button.
but I check the existing errors, it seems that not all the errors need this button. because some errors are only a warning message to users. We don't want them to save log and send to Ledger support. I leave the possibility to hide the button with this variable. 
I don't know whether I should config all the old errors so that we remove this button for certain errors? if yes, I need someone else to help me with that

src/log-report.js
hzheng-ledger: fixed

src/components/GenericErrorView.js
hzheng-ledger: I've talked with @cthepot-ledger offline. We decided to refer to the current logic of ledger-live-desktop app.
I found that in LLD, the "save logs" button always appears.

refer to:
https://github.com/LedgerHQ/ledger-live-desktop/blob/766fdf662b78d1d9824367687a5ffe725feef730/src/renderer/components/DeviceAction/index.js#L177
and
https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/components/DeviceAction/rendering.js#L344

Screen shot from LLD:
Screenshot 2021-04-12 at 15 08 40

@gre can you confirm that?

So, I assume that the my current work has the same behaviour of LLD and we can merge this PR after review.

gre: yes it always appears for unhandled errors like this one. so yeah to me we can be iso.

cthepot-ledger:
assigning @nabil-brn and tagging QA based on this [comment](https://github.com/LedgerHQ/ledger-live-mobile/pull/1654#discussion_r612229232)

Arnaud97234 (approved):

gre (approved):

PR #1661, opened 2021-04-08
LL-2540 All time support

JunichiSugiura:
### Type

Feature

### Context

LL-2540

- Portfolio Graph
- Account Graph
- Asset Distribution


JunichiSugiura:
Remaining flow type errors are mainly due to `BalanceHistoryData` that has `data` property of `?Date`.

PR #1663, opened 2021-04-13
LL-5074 Change wording for wrong device error

haammar-ledger:
Context: https://ledgerhq.atlassian.net/browse/LL-5074


gre (commented):
not a problem for now but note for future: we should only update the /en lang because others are generated and it could conflicts with crowdin

haammar-ledger:
I knew that I knew it... but wasn't sure. I can revert the others no prob.

gre (approved):

PR #1668, opened 2021-04-16
LL-5161 Upgrade to RN 0.64

JunichiSugiura:
Upgrade react-native and other native dependencies.

### Type

Dependency upgrade

Pretty much entire app is affected but here're more details.
https://github.com/LedgerHQ/ledger-live-mobile/pull/1682#issue-624247202


gre:
@JunichiSugiura  to be discussed on monday: I think it would be wise to attempt to also include #1719 in this rework. let's give a try at least and check all is fine (specifically no issue when being on different locales).

JunichiSugiura:
Note:
[ ] RNCSafeAreaProvider error

PR #1669, opened 2021-04-16
fix error wording for different device/platform

hzheng-ledger:
Jira ticket
https://ledgerhq.atlassian.net/browse/LL-5221

Android Nano S

Screenshot 2021-04-16 at 15 14 08

Android Nano X

Screenshot 2021-04-16 at 15 15 08

IOS Nano S (For Nano X, it is similar)

Screenshot 2021-04-16 at 15 20 09

### DIFF
--- a/src/components/TranslatedError.js
+++ b/src/components/TranslatedError.js
@@ -1,5 +1,6 @@
 // @flow
 import { useTranslation } from "react-i18next";
+import { Platform } from "react-native";
 
 type Props = {
   error: ?Error,
@@ -18,12 +19,21 @@ export default function TranslatedError({ error, field = "title" }: Props) {
     return null;
   }
   // $FlowFixMe
-  const arg: Object = { message: error.message, ...error };
+  const arg: Object = { message: error.message, returnObjects: true, ...error };
   if (error.name) {
-    const translation = t(`errors.${error.name}.${field}`, arg);
+    let translation = t(`errors.${error.name}.${field}`, arg);
     if (translation !== `errors.${error.name}.${field}`) {
-      // It is translated
-      return translation;
+      if ((typeof translation === "object") && ("productName" in arg)) { // it has specific translation for different device and platform
+        const platform = Platform.OS;
+        const device = arg.productName.includes("Nano S")? "nanoS" : "nanoX";
+        translation = t(`errors.${error.name}.${field}.${platform}.${device}`, arg);
+        if (translation !== `errors.${error.name}.${field}.${platform}.${device}`) {
+          return translation;
+        }
+      }
+      else if (typeof translation === "string") {
+        return translation;
+      }
     }
   }
   return t(`errors.generic.${field}`, arg);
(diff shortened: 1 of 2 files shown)


src/components/TranslatedError.js
```diff
@@ -18,12 +19,26 @@ export default function TranslatedError({ error, field = "title" }: Props) {
     return null;
   }
   // $FlowFixMe
-  const arg: Object = { message: error.message, ...error };
+  const arg: Object = { message: error.message, returnObjects: true, ...error };
```
henri-ly, line 22: I'm not sure what does `returnObjects` ?

hzheng-ledger: by default , returnObjects= false.
the transition accepts only string. This was the old behaviour. 
Now, I have android/ios, nano s/x for different translation, so the translation is an object (an object contains keys like "ios", "android"....) You can refer to my "common.json " to understand how it works.
Thanks

src/locales/en/common.json
```diff
@@ -105,11 +105,29 @@
     },
     "ConnectAppTimeout": {
       "title": "Sorry, no device found",
-      "description": "Please make sure your {{productName}} is unlocked with Bluetooth enabled."
+      "description": {
+        "ios": {
+          "nanoS": "Please make sure your {{productName}} is unlocked and has Bluetooth enabled.",
```
LFBarreto, line 110: this case seems odd :sweat_smile: nanoS can't be connected via bluetooth. we can keep it

src/components/TranslatedError.js
```diff
@@ -18,12 +19,26 @@ export default function TranslatedError({ error, field = "title" }: Props) {
...
+    let translation = t(`errors.${error.name}.${field}`, arg);
     if (translation !== `errors.${error.name}.${field}`) {
-      // It is translated
-      return translation;
+      if (typeof translation === "object" && "productName" in arg) {
+        // it has specific translation for different device and platform
+        const platform = Platform.OS;
+        const device = arg.productName.includes("Nano S") ? "nanoS" : "nanoX";
+        translation = t(
+          `errors.${error.name}.${field}.${platform}.${device}`,
```
LFBarreto, line 31: maybe put these string id in a const since they are reused to compare afterwards

LFBarreto (changes requested):
I think we could have simplified this by being more generic in the message rather than hacking the specifics.
maybe "Please make sure your {{productName}} is unlocked and connected to your phone." :shrug: 
@dasilvarosa ?

src/locales/en/common.json
juan-cortes: there'd never be a case though, why keep the wording ? 🤔

juan-cortes (approved):
I'm not sure about the approach of the wording, nanoS will not be connected to iOS (unless through a proxy) and the wording about having Bluetooth doesn't make sense in that case. Other than that, if QA is happy, I'm ok.

achabrolin-ledger:
FYI: `yarn flow` fails with 912 errors, unrelated with modified code. Don't know how nor why it fails

PR #1675, opened 2021-04-20
fix(Alert): set width 100% always - LL-5255

ghost:
### Type

UI Fix

### Context

On Swap, info alert is not full width, content is not visible.

Instead of wrapping all <Alert> in a <View> with width=100%, i fixed the Alert component to always have 100% width (there is no use-case were Alert does not take all available horizontal space).
I updated the SkipDeviceVerification component that used the wrapping solution.

Issue: https://ledgerhq.atlassian.net/browse/LL-5255

Also, alert in SkipDeviceConfirmation after Buy verification without device should not change.

### DIFF
--- a/src/components/Alert.js
+++ b/src/components/Alert.js
@@ -222,7 +222,6 @@ export default function Alert({
       style={[
         styles.root,
         borderColor && styles.withBorder,
-
         {
           backgroundColor,
           borderColor: borderColor || "transparent",
@@ -265,6 +264,7 @@ export default function Alert({
 
 const styles = StyleSheet.create({
   root: {
+    width: "100%",
     flexDirection: "column",
     borderRadius: 4,
     paddingVertical: 16,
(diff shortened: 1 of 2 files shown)


machard (approved):

juan-cortes (dismissed):

Arnaud97234:
@rjarasson-ledger Your fix works but by testing it I saw the same error on the confirm swap drawer. Could you fix it in the same PR or you prefer me to open a new bug task in Jira?
- Select accounts and amount
- Swap summary
- Drawer when waiting from user to confirm on device

IMG_1201

gre (dismissed):
:go:

rjarasson-ledger:
I tried to reproduce the issue, but unseccessfully.
On my side, on both Android and iOS, the issue was present, but fixed by this PR.
As i am not able to reach this specific screen, i added the Alert in other DeviceAction confirm drawers like Receive or Allow Manager.

rjarasson-ledger:
@Arnaud97234 I was not able to find a specific fix to this drawer.

But https://github.com/LedgerHQ/ledger-live-mobile/pull/1700 should fix both the Error to display correctly, and the Alert component to be fixed by the current PR.

It seems like this drawer on Swap is completely buggy, and provoke those side effects.

This PR fixes the Alerts not affected by other UI glitches.

Arnaud97234 (approved):
To test this PR I had to pull https://github.com/LedgerHQ/ledger-live-mobile/pull/1700 in it.
Both are ok. ✅

PR #1679, opened 2021-04-22
LL-4614 Inline installation on connectApp flows as experimental

juan-cortes:
https://ledgerhq.atlassian.net/browse/LL-4614

This functionality allows the user to install an application when, during a device flow, we detect that they don't have such application installed. This affects the Send, Receive, and Add account flows and enables the user to complete the flow without having to leave the modal/screen, go to the manager, install the application, and return.

This means that we can detach a bit more the user's awareness of what an app is, and what apps are installed on their device, minimise the friction on the flows too.

It's controlled by a new ENV variable `EXPERIMENTAL_INLINE_INSTALL` since we should thoroughly test this before releasing. It can also be enabled in the experimental section of the settings.

### How to test this
- All flows of the app should work with and without the experimental setting.
- When the experimental setting is ON, it should solve app dependencies in-flow
- When the experimental setting is OFF, it should refer the user to the manager


gre (approved):

gre:
I'm flagging this as important because it's merged on LLD but not yet LLM.

Arnaud97234 (changes requested):
### Bug found
@juan-cortes 
User shouldn't be able to close the installation drawer while an app is installing:
- Remove close button
- Remove closing drawer when tapping outside
IMG_1194
IMG_1193

juan-cortes:
Indeed it should be blocking on these steps I guess

juan-cortes:
I'm taking a look today

juan-cortes:
@Arnaud97234 after talking with @gre and since even if the user leaves mid install (device is in processing) it's not locked in that state, since this is experimental, it's not a blocker for merging. I did have to update the PR because the experimental setting wasn't working on my end. Please test again

gre:
we have to hodl the PR and merge live-common pr first because there were breaking change in live-common that needed some fixes (like shipping fees).
we can merge after it.

gre:
included in #1681

PR #1685, opened 2021-04-28
LL-5284 Providers list entry for swap

juan-cortes:
Pushing this early for @haammar-ledger  to see if the 533d57f commit could be useful for cherry-picking


juan-cortes:
This is no longer needed, it will be part of a more up-to-date PR for the new wyre integration.

PR #1700, opened 2021-05-06
LL-5522 - fix(Swap): Confirmation bottom modal fit the DeviceActionModal markup…

ghost:
### Type

UI Fix

### Context

JIRA: https://ledgerhq.atlassian.net/browse/LL-5522

The Confirmation bottom modal on Swap uses a DeviceAction rendering, but does not match the same markup as the DeviceActionModal used anywhere else.
Due to current rendering, device error is displayed wrongly, the paddings mismatch other modals, and the alert component cannot fit full width.

I completely removed the SafeAreaView and re-used the ModalBottomAction component the same way as in the DeviceActionModal.

If should also allow https://github.com/LedgerHQ/ledger-live-mobile/pull/1675 to work correctly

### DIFF
--- a/src/screens/Swap/Confirmation.js
+++ b/src/screens/Swap/Confirmation.js
@@ -1,9 +1,8 @@
 // @flow
 
 import React, { useCallback, useEffect, useState } from "react";
-import { StyleSheet } from "react-native";
+import { StyleSheet, View } from "react-native";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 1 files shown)


valpinkman (approved):

Arnaud97234 (approved):
To test this PR I had to pull https://github.com/LedgerHQ/ledger-live-mobile/pull/1675.
Both are ok.
✅

PR #1704, opened 2021-05-07
fixes import of rxjs for future 7.x

gre:
if it passes CI we should be good.


valpinkman (approved):

gre:
we'll recheck this more carefully yes; in any case these changes can gets in even with 6.x

PR #1707, opened 2021-05-10
LL-5429 Rename exported logs file

juan-cortes:
Added a more meaningful name for the exported logs on LLM, note that this does not work for iOS since there's a limitation on the library we use so it will continue to default to `file.json` for that platform. On Android however, it will now follow a naming format such as `ledger-live-mob-2.26.0-(arm64-v8a 156)-2021-05-10-logs.json` which should be a lot more meaningful.

> I noticed while looking at the logs that the output was not a valid JSON string so I fixed that, but still, the exported logs are not as useful as the desktop ones and we should add more information to them such as the number of accounts, etc. Perhaps this is a good next iteration to consider cc @gre @cthepot-ledger 

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-5429

Since triggering the generic error component was a pita, I've added a trigger to the Debug Crash screen on the hidden debug menu, which allows QA to render the component without triggering an actual error. Just go to the `Debug menu ⇨ Mock & Test ⇨ Debug crash` and trigger it from there.

image

### DIFF
--- a/src/components/GenericErrorView.js
+++ b/src/components/GenericErrorView.js
@@ -14,7 +14,6 @@ import logReport from "../log-report";
 import Button from "./Button";
 import DownloadFileIcon from "../icons/DownloadFile";
 import cleanBuildVersion from "../logic/cleanBuildVersion";
-import { localeIds } from "../languages";
 
 class GenericErrorView extends PureComponent<{
   error: Error,
--- a/src/components/GenericErrorView.js
+++ b/src/components/GenericErrorView.js
@@ -4,6 +4,7 @@ import { View, StyleSheet } from "react-native";
 
 import Share from "react-native-share";
 import { Trans } from "react-i18next";
+import VersionNumber from "react-native-version-number";
 import LText from "./LText";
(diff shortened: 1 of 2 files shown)


hzheng-ledger (dismissed):
looks good to me

ghost (dismissed):

Arnaud97234 (changes requested):
@juan-cortes 
See comment in: https://ledgerhq.atlassian.net/browse/LL-5429

On iOS the file name is the default one "file.json"

juan-cortes:
I mentioned in the original message that iOS files could not be modified, it's a limitation of the library we use. We could rework the whole export so that it creates a new temporary file with a given name (maybe) but it's out of the scope of this task, to me, it improves the situation of android so it's a merge.

cc @gre up to core

juan-cortes:
Got the OK from @cthepot-ledger for the iOS limitation, I'm assigning him because I don't know who to assign otherwise

Arnaud97234 (dismissed):

valpinkman (dismissed):

PR #1711, opened 2021-05-12
LL-5273 Allow access to swap feature without the device wall

juan-cortes:
When accessing the swap feature we currently require users to have the exchange app installed and up to date and when inside the form we limit the selectors depending on whether they have the applications installed or not. This means we are capping the feature before users can access it. This feature allows the user to enter the swap form and request a form without knowing that they need the exchange app and, once the inline install is no longer experimental, they will be able to fulfill that requirement without leaving the form.

### Type
Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-5273

- Without experimental inline install, try cases with/without all required apps/accounts
- With experimental inline install, try cases with/without all required apps/accounts

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@ledgerhq/hw-app-xrp": "5.51.1",
     "@ledgerhq/hw-transport": "5.51.1",
     "@ledgerhq/hw-transport-http": "5.51.1",
-    "@ledgerhq/live-common": "19.9.0",
+    "@ledgerhq/live-common": "19.9.0-beta.2",
     "@ledgerhq/logs": "5.50.0",
     "@ledgerhq/react-native-hid": "5.51.1",
     "@ledgerhq/react-native-hw-transport-ble": "5.51.1",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@ledgerhq/hw-app-xrp": "5.51.1",
     "@ledgerhq/hw-transport": "5.51.1",
     "@ledgerhq/hw-transport-http": "5.51.1",
-    "@ledgerhq/live-common": "19.9.0-beta.2",
+    "@ledgerhq/live-common": "19.9.1",
     "@ledgerhq/logs": "5.50.0",
     "@ledgerhq/react-native-hid": "5.51.1",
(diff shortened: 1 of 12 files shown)


Arnaud97234 (changes requested):
@juan-cortes Metro bundler throw an error when buiding the application.
image (1)

juan-cortes:
Ready when you are, @Arnaud97234

JunichiSugiura (dismissed):

Arnaud97234 (changes requested):
On Android everything is OK.

On iOS, After selecting the device (summary step):
- If all required app are present on the device: The drawer asking the user to open the exchange app is missing. After opening the exchange app, I can confirm the swap on the device but nothing is displayed in LL.
- If some app are missing on the device nothing happen. We stay on the swap summary.

**Expected**
expected_swap

**Actual**
actual_Swap

juan-cortes:
iOS, I've slowly developed a hatred towards you, deep, rooted, and unavoidable.

valpinkman (dismissed):

juan-cortes:
I can't currently test due to the super high limits on swap, I'll test tomorrow again

juan-cortes:
There is a limitation on react-native-modal, we bumped into this in the past but I must've forgotten about it. We need to wait for one modal to close completely before we can open another one. This is only affecting iOS, hence the bug found by @Arnaud97234 and I left a note so we don't delete that code. In any case, it should be ready for QA now, and I guess also a new dev review @valpinkman since I had to add some things. :sorry:

juan-cortes:
Very strange that this is a new bug, are there any bugs without experimental install? To me it's two different things, this would be a bug on inline install, which is still experimental :thinking:

juan-cortes:
With the latest version any error would close the bottom modal and kill the flow, this is to bypass the fact that we can't lock/unlock in the middle of an action, the current version leaves the user stuck when there's an error and they have no way out. This incidentally fixes that case too.

I believe this should also fix the bug that you were experiencing @Arnaud97234 let me know if you have time to check it.

Arnaud97234 (approved):
Tested on iOS and Android.
Tested with Floating and fixed rates.
good to merge ✅

gre (approved):

PR #1714, opened 2021-05-17
fix delegation counter value for tezos

hzheng-ledger:
Fix https://ledgerhq.atlassian.net/browse/LL-5208
the counter value was wrong before because the tezos price was calculated based on the delegation operation date


valpinkman (changes requested):
you can just use `new Date()`

hzheng-ledger:
> you can just use `new Date()`

Fixed. thanks

gre (changes requested):
sorry to come back now, but actually we shouldn't set `date` prop at all. date is optional and if it's not set, it's actually doing "latest".
It's better to do this because new Date() always will change ref and therefore always retrigger a re-rendering vs not setting a prop will be only updating when the countervalues are changing.

hzheng-ledger:
> sorry to come back now, but actually we shouldn't set `date` prop at all. date is optional and if it's not set, it's actually doing "latest".
> It's better to do this because new Date() always will change ref and therefore always retrigger a re-rendering vs not setting a prop will be only updating when the countervalues are changing.

Thanks for the feedback. Fixed now

ghost (approved):
Ready for QA imho

ghost (approved):

nabil-brn (approved):

PR #1719, opened 2021-05-20
Enable Hermes engine for Android

JunichiSugiura:
### Type
Internal Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-5764

we have to validate there are no regression but it should improve performance.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -87,7 +87,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    enableHermes: false,  // clean and rebuild if changing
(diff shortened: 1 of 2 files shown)


LFBarreto (approved):

gre:
we need a pre evaluation from a dev that there are no obvious regression. i'm thinking specifically about formatting of numbers and dates on different phone languages / locales you set (possibly you need to make a custom build to enable langs to test this btw!)

valpinkman (approved):

gre:
i'm closing this towards the idea to unify this work and put it inside #1668

PR #1721, opened 2021-05-21
Use Fixed rate by default instead of Floating rate

hzheng-ledger:
In swap screen, we should use Fixed rate by default instead of Floating rate
https://ledgerhq.atlassian.net/browse/LL-5259

### DIFF
--- a/src/screens/Swap/Form/Amount.js
+++ b/src/screens/Swap/Form/Amount.js
@@ -61,7 +61,7 @@ const SwapFormAmount = ({ navigation, route }: Props) => {
     [fromCurrency, selectableCurrencies, toCurrency],
   );
   const [tradeMethod, setTradeMethod] = useState<"fixed" | "float">(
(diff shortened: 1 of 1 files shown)


IAmMorrow (approved):

cthepot-ledger:
not it's floating rate by default. the inactive UI component should be floating rate by default as well.

cthepot-ledger:
should not merge this

hzheng-ledger:
@cthepot-ledger 
Hi, Chris, I just checked the mobile application UI in swap page.
Screenshot 2021-06-08 at 10 31 30

Since the "Fixed rate"/"Floating rate" selection is after the coin selection. So, the "Fixed rate"/"Floating rate" CTA is always active. I assume that nothing needs to be done for this ticket. Could you please confirm that? Thanks

gre:
on Jira, ticket is abandoned. closing this PR unless some mistake? we can reopen if there are.

PR #1726, opened 2021-05-28
LL-4998 [Polkadot] Add a flow to change controller

ghost:
In the specific situation where the current account is a stash with a different controller, provide a button to the user to set the account as its own controller.

**Requires:** LedgerHQ/ledger-live-common#1217
**Requires:**  minimum Nano app v7.9050.0

**How to test:** from a new account, do a first bond that uses a different account as controller (can only be done using https://polkadot.js.org/apps)

Screenshot_1626789701


gre (changes requested):
CI fails. conflicts with upstream.

haammar-ledger:
Blocked by https://ledgerhq.atlassian.net/browse/NAPP-88: need a new Nano app version for this to work.

gre (approved):

PR #1731, opened 2021-05-31
BLE performance optimisations

gre:
This adds back the logic to pause the background synchronisation of accounts during critical phases like Manager or device flows.

It also includes this work: https://github.com/LedgerHQ/ledger-live-common/releases/tag/v19.11.0


juan-cortes (changes requested):
While this would fix instability issues in some cases, it would not fix it for swap since it seems to be using `ModalBottomAction` and not `DeviceActionModal`. We either need to add it to that too (via a prop or something) or refactor the confirmation for swap so that it uses `DeviceActionModal`

The reason it's not using a simple `DeviceActionModal` is that for swap confirmation we are chaining several things, the initSwap action and the sign, and broadcast parts.

gre:
@juan-cortes should be covered now

juan-cortes:
> @juan-cortes should be covered now

Awesome

juan-cortes (dismissed):
🚀

nabil-brn (changes requested):
I don't know if it's on develop branch but the experimental header is displayed on the drawers.

Screenshot_20210531-180755_LL  STAGING
---
6952f2727 (HEAD -> LL-5685, upstream/LL-5685) update live-common

Arnaud97234 (changes requested):
See comment in https://ledgerhq.atlassian.net/browse/LL-5781

JunichiSugiura (commented):
LGTM

Arnaud97234:
Last remaining topic for BLE disconnection: https://ledgerhq.atlassian.net/browse/LL-5801/ It's not a blocker. We can merge

Arnaud97234 (approved):

MortalKastor (approved):

valpinkman (approved):

LFBarreto (approved):

PR #1732, opened 2021-05-31
LL-5559 - fix android graph crash

machard:
https://ledgerhq.atlassian.net/browse/LL-5559

I replaced d.value by NaN and it reproduced the same stacktrace as on sentry. So I supposed I targeted the correct lines.
However it feels like I am not fixing the root cause (having NaN/null as values). but at least it fix the crash.


LFBarreto (approved):
:+1: :confetti_ball:

gre:
let's take this as part of our next build. no way to test, we'll just confirm all is fine on our next release.

PR #1743, opened 2021-06-01
(ReactNavigation): enable react native screens optimizations

LFBarreto:
enable react native screens as per https://reactnavigation.org/docs/react-native-screens
should optimize some rendering in heavy navigation stack pages

### Type

Code Quality Improvement

Heavy stack pages navigation


gre:
HODL. this library is tricky, we had issue in the past with it. i would suggest to hodl and let's create a task in our jira.

gre:
I'll close for now, we can rediscuss / schedule task at least.

PR #1744, opened 2021-06-02
Align prettier rules to LLD for 100 line breaks

juan-cortes:
We have different rules for prettier on LLD and LLM, and since I try to reuse as much code as possible, this ends up being a pita when the code doesn't look the same. I don't mind if we align LLD to LLM or the other way around, but please one single format for styles.

### Type

Code Quality Improvement

Nothing to test


valpinkman (approved):

gre:
HODL because it's the worse timing to do this now. hopefully after we have a release ;)

juan-cortes:
> HODL because it's the worse timing to do this now. hopefully after we have a release ;)

Core team's call when, but we should align at some point

gre:
it's one of these hard to do PR which impact everywhere. it sounds like we should merge as much PR before as possible and possibly even upgrade prettier (i think we are not on latest).

I actually wonder if we shouldn't just wait TypeScript rework and do it at this time.

I'll close the PR for now and it will need to be added as sub task of TS LLM.

PR #1745, opened 2021-06-02
LL-5281  Swap with Wyre

juan-cortes:
### Description
The functionality is only partial since we still have dependencies for both KYC validation and actual swap completion, however, it should work up until the exchange rates and crash on the summary for Wyre.

This pr also includes a rework on the app state regarding swap making it more manageable as well as general sync between LLD and LLM naming and component logic, reusing as much as possible unlike the previous versions (my bad).

It is rebased with develop as of 29th of Jun, which includes dropping the provider selector we introduced on v0 for parasap. Artifacts from this pr should be testable, but require a few env to work. First of all, we need to launch the app pointing to staging v3 `SWAP_API_BASE=https://swap-stg.ledger.com/v3` and in case we want to disable a provider (it will fallback to changelly if both are available) we have a new env `SWAP_DISABLED_PROVIDERS` which can be used to disable one or more providers.

For instance if we launch with `SWAP_DISABLED_PROVIDERS="wyre,changelly"` we can test the no providers case.
To show the debug menu on the providers screen, pass `DEBUG_PLATFORM=1` at launch/env

There's a new screen in the debug menu to allow easy resetting of the KYC status and specific providers.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-5281

TODO


cthepot-ledger:
@juan-cortes to add LL-5955

LFBarreto (approved):

gre:
good to go. i'll merge the live-common activation on a separate PR .

PR #1758, opened 2021-06-08
LL-3449 (Swap/History) Add warning about swaps not being synced

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3449

-  Launch LLM on data that does have swap history, compare to the screenshot.
-  It should not show if there are no operations.

### DIFF
--- a/ios/ledgerlivemobile.xcodeproj/project.pbxproj
+++ b/ios/ledgerlivemobile.xcodeproj/project.pbxproj
@@ -300,7 +300,7 @@
 				00DD1BFF1BD5951E006B06BC /* Bundle React Native code and images */,
 				A3B86A0A04E74A989A896DF8 /* Upload Debug Symbols to Sentry */,
 				760D57A3243689E700E1EDF6 /* ShellScript */,
-				DF1473EF7BC55974872D9970 /* [CP] Embed Pods Frameworks */,
+				CF86CBF56A06289606586AED /* [CP] Embed Pods Frameworks */,
 			);
 			buildRules = (
 			);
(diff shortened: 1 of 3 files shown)


src/locales/en/common.json
```diff
@@ -1937,6 +1937,7 @@
       },
       "history": {
         "tab": "History",
+        "disclaimer": "Your swap desktop transactions are not synchronized with Ledger Live Desktop",
```
haammar-ledger, line 1940: You probably mean "mobile" at least once :smiley: but not sure which one.

src/screens/Swap/FormOrHistory/History/index.js
```diff
@@ -122,6 +123,16 @@ const History = () => {
   return (
     <View style={[styles.root, { backgroundColor: colors.background }]}>
       <TrackScreen category="Swap" name="History" />
+      <View
+        style={[
+          styles.alertWrapper,
+          { backgroundColor: colors.white, borderColor: colors.lightFog },
```
haammar-ledger, line 129: Why not include this in `styles.alertWrapper`?

haammar-ledger (changes requested):

src/locales/en/common.json
juan-cortes: Hahahaha indeed, nice catch 🤦🏼

ios/ledgerlivemobile.xcodeproj/project.pbxproj
```diff
@@ -300,7 +300,7 @@
 				00DD1BFF1BD5951E006B06BC /* Bundle React Native code and images */,
 				A3B86A0A04E74A989A896DF8 /* Upload Debug Symbols to Sentry */,
 				760D57A3243689E700E1EDF6 /* ShellScript */,
-				DF1473EF7BC55974872D9970 /* [CP] Embed Pods Frameworks */,
+				CF86CBF56A06289606586AED /* [CP] Embed Pods Frameworks */,
```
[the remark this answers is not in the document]
juan-cortes: This happens when you run yarn pod, which I needed to run for the build to work, so as far as I'm concerned it's needed. In any case, I also have no idea what it means, a hash of something I guess.

src/screens/Swap/FormOrHistory/History/index.js
juan-cortes: Because colors comes from a hook and changes per theme, it's not available on the styles as far as I know 🤔

haammar-ledger (dismissed):
:ship:

Arnaud97234 (changes requested):
@juan-cortes see my comment in https://ledgerhq.atlassian.net/browse/LL-3449.

The info box is present even if the swap history is empty. If there is no operation in the history the info box shouldn't be displayed.

juan-cortes:
Good catch @Arnaud97234 I will make it only visible when we have swaps, like on LLD

juan-cortes:
Ready for you @Arnaud97234

Arnaud97234 (approved):

Arnaud97234 (commented):
Tested on both Android and iOS.
- If operation history -> info box
- If no operation history -> No info box
✅

PR #1761, opened 2021-06-09
feat(Swap): update success screen wording and UI - LL-5733

rjarasson-ledger:
### Type

UI Polish

### Context

This PR updates the Swap Confirmation screen to show better that broadcast is successful but operation is still pending.
Icon updated with Clock, wording updated, and Swap ID filled in the Disclaimer alert.

JIRA: [LL-5733](https://ledgerhq.atlassian.net/browse/LL-5733)

At the end of a Successful Swap, should look like this:

**NOTE: Not tested with a REAL SWAP (i wanted to avoid burning fees since QA should test it for real).**
The part that i have not tested is if the targetCurrency is correctly filled in the description text.

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -1837,10 +1837,10 @@
         "cta": "Close"
       },
       "pendingOperation": {
-        "description": "Your Swap operation has been sent to the network for confirmation. It may take up to an hour before you receive your swapped crypto assets.",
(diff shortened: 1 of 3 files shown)


juan-cortes (approved):

LFBarreto (approved):

Arnaud97234 (approved):
Tested on both iOS and Android.
✅

PR #1762, opened 2021-06-10
LL-5914 (AuthPass): fix some issue with navigation hooks imported outside of navigation context

LFBarreto:
(AuthPass): fix some issue with navigation hooks imported outside of navigation context
### Type

Bug Fix
### Context

[LL-5914]

Password lock screen should no longer crash

[LL-5914]: https://ledgerhq.atlassian.net/browse/LL-5914

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -75,7 +75,7 @@ function ButtonWrapped(props: BaseButtonProps) {
   const isFocused = useIsFocused(); // @Warning be careful not to import the wrapped button outside of navigation context
   const useTouchable = useContext(ButtonUseTouchable);
   return (
-    <Button
+    <BaseButton
       {...props}
       useTouchable={useTouchable}
       colors={colors}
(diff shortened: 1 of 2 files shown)


valpinkman (approved):

haammar-ledger (dismissed):
Maybe the best way to avoid importing the wrong version would be to remove the default export and make both `Button` and `WrappedButton` explicit exports?
Not sure this is possible with `memo<BaseButtonProps>` though...

LFBarreto:
> Maybe the best way to avoid importing the wrong version would be to remove the default export and make both `Button` and `WrappedButton` explicit exports?
> Not sure this is possible with `memo<BaseButtonProps>` though...

i renamed the export from Button to BaseButton it should do the trick for now :+1:

haammar-ledger (dismissed):
One can still do
`import Button from "../../components/Button";`
by mistake, but anyway this PR fixes the current issue :+1:

valpinkman (approved):

juan-cortes (approved):

Arnaud97234 (approved):
No crash when editing the input field on password lock screen. 
Tested on iOS and Android.
✅

PR #1765, opened 2021-06-11
fix(Swap): drop the concept of payoutNetworkFees from UI - LL-5931

rjarasson-ledger:
> This pr is tightly linked to LedgerHQ/ledger-live-common#1244 and they should land on production together. Delivering one of them independently from the other will make the amounts not match for the users doing swaps.

### Type

Feature

### Context

Originally when we delivered the floating rate trading method for swap there was a new concept of payoutNetworkFees, which represents an amount that is charged as a fee on top of the rate provided. We needed to display this because the amount on the device included that amount, but the amount to be received had it deducted.

With the latest code on staging, this is no longer the case and the binary payload amount (which we use on the device verification) deducts this amount, allowing us to drop the UI that referred to this concept altogether.

DEPENDS ON LIVE-COMMON: https://github.com/LedgerHQ/ledger-live-common/pull/1244
LLD PR: https://github.com/LedgerHQ/ledger-live-desktop/pull/3899

* Use yalc version of LLC with branch from PR https://github.com/LedgerHQ/ledger-live-common/pull/1244
* Use .env variable: ```SWAP_API_BASE=https://swap-stg.ledger.com/v2```
* Start swap with floating rate, go to sign step (no need to accept swap)
* Only amount is displayed in the summary screen, and should match the `get estimated` value on device

### DIFF
--- a/src/screens/Swap/FormOrHistory/Form/Summary/SummaryBody.js
+++ b/src/screens/Swap/FormOrHistory/Form/Summary/SummaryBody.js
@@ -20,7 +20,6 @@ import {
 
 import LText from "../../../../../components/LText";
 import CurrencyUnitValue from "../../../../../components/CurrencyUnitValue";
-import InfoModal from "../../../../../components/InfoModal";
 import SectionSeparator, {
   ArrowDownCircle,
 } from "../../../../../components/SectionSeparator";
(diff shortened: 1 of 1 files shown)


rjarasson-ledger:
cc @juan-cortes

juan-cortes (approved):
lgtm

cthepot-ledger:
@gre already approved on jira so it should be approved here to

PR #1777, opened 2021-06-15
LL-5905 (Swap) Remove the bifurcation screen

juan-cortes:
Should be a good starting point for both the catalog feat and the rework for wyre swap.

### Type
Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-5905

Access swap, there should be no questions, just form (loading/ip/etc) for changelly


rjarasson-ledger:
Included and supereded by https://github.com/LedgerHQ/ledger-live-mobile/pull/1785

PR #1779, opened 2021-06-16
LL-6018 LL-5808 (ServiceStatusProvider): use currency filtered incident hook

LFBarreto:
(ServiceStatusProvider): use currency filtered incident hook
### Type

Feature
### Context

[LL-6018]

Status incidents should now be filtered according to user available currencies (accounts added)

[LL-6018]: https://ledgerhq.atlassian.net/browse/LL-6018


achabrolin-ledger (dismissed):

ychen-ledger (commented):
Validated with: 

- iOS 14.6
- LLM 2.33.0
- BTC/ETH/XRP

PR #1785, opened 2021-06-18
LL-5399 - Platform Catalog

rjarasson-ledger:
Adds the Platform Catalog "Browse".
It also implements bifurcation screens removal, and analytics.

Depends on: https://github.com/LedgerHQ/ledger-live-common/pull/1266

### Type

Feature

### Context

The Platform catalog uses Live Common's CatalogProvider to display a list of apps available in LLM.
You can enable "debug" apps by using ENV variable `PLATFORM_DEBUG`.

JIRA ticket: https://ledgerhq.atlassian.net/browse/LL-5399
Subtasks:
* https://ledgerhq.atlassian.net/browse/LL-5905
* https://ledgerhq.atlassian.net/browse/LL-5906
* https://ledgerhq.atlassian.net/browse/LL-5601
* https://ledgerhq.atlassian.net/browse/LL-6039

### TODO

* [x] remove bifurcation screens
* [x] implement new Catalog
* [x] add Platform Analytics
* [x] use translated contents
* [ ] final wording and banner illustration
* [x] COMING SOON Cards
* [ ] MAYBE - More app info modal (ℹ️ icon)
* [x] Rebase on https://github.com/LedgerHQ/ledger-live-mobile/pull/1777 to remove previous integration
* [ ] MAYBE - Add an experimental switch for displaying experimental apps (useless until catalog is remotely fetched)
* [ ] ~CANCELLED - Add in-between drawer to let user accept it is an external third-party feature, maybe TOS ?~

* New "Browse" section in SideBar
* Catalog with 2 apps.- ParaSwap & Wyre (3 if PLATFORM_DEBUG)
* CTA for Poll
* CTA for accessing dev portal


src/screens/Platform/CatalogCTA.js
```diff
@@ -0,0 +1,103 @@
...
+          </View>
+          <LText style={styles.description}>{description}</LText>
+        </View>
+        <IconExternalLink size={18} color={colors.smoke} />
+      </View>
+    </TouchableOpacity>
+  );
+};
+
+const getStyles = (colors, type) =>
```
LFBarreto, line 53: does this work in the end? if you switch theme and go back does it change here too?

src/screens/Platform/CatalogCTA.js
```diff
@@ -0,0 +1,103 @@
...
+      marginBottom: 16,
+    },
+    iconContainer: {
+      marginRight: 8,
+    },
+    title: {
+      fontSize: 18,
+
+      ...(type === "secondary"
+        ? { color: colors.text }
```
LFBarreto, line 88: works like that but in this case since you need ternary for color only why not do:
`colors: type === "secondary" ? ... :  ...
`

LFBarreto (approved):

src/screens/Swap/FormOrHistory/Form/index.js
```diff
@@ -134,7 +133,9 @@ const Form = ({
     <View style={styles.root}>
       <TrackScreen category="Swap" name="Form" />
       <Alert
+        id="transfer.swap.form.paraswapCta"
         type="hint"
+        closeable={true}
```
rjarasson-ledger, line 138: remove the ={true}
Just `closeable` is shorthand usage that we should use for boolean.

src/components/Alert.js
```diff
@@ -217,48 +240,67 @@ export default function Alert({
...
-        },
-      ]}
-    >
-      <View style={[styles.container, vertical && styles.vertical]}>
-        {left || (!noIcon && icon) ? (
-          <View style={vertical ? styles.topContainer : styles.leftContainer}>
-            {left || icon}
-          </View>
-        ) : null}
+    !dismissedBanners.includes(id) && (
```
rjarasson-ledger, line 250: Add a useMemo for "isDissmissed" here.

src/screens/Platform/CatalogCTA.js
rjarasson-ledger: You are right, i think i just wanted to use the same approache for all the variants. But it's not as beautiful as it could be anyway.

Plus there is a typo some lines after in "colors:".
We should fix it.

src/components/Alert.js
```diff
@@ -306,4 +353,9 @@ const styles = StyleSheet.create({
     textDecorationLine: "underline",
     marginTop: 8,
   },
+  closeContainer: {
+    position: "absolute",
+    top: "-20%",
```
LFBarreto, line 358: i'd rather use px values instead of % for positioning

LFBarreto (changes requested):

LFBarreto (approved):

PR #1791, opened 2021-06-24
LL-5446 (Compound): disable lending entries for live with non enabled accounts

LFBarreto:
(Compound): disable lending entries for live with non enabled accounts
### Type

Feature
### Context

[LL-5446]

Compound entry points for accounts should be limited to accounts that have it already enabled.
Also the global entry is shown to apps having at least one lending enabled account.

[LL-5446]: https://ledgerhq.atlassian.net/browse/LL-5446

### DIFF
--- a/src/modals/Create.js
+++ b/src/modals/Create.js
@@ -6,7 +6,10 @@ import { useTranslation } from "react-i18next";
 import { useSelector } from "react-redux";
 
 import { NavigatorName, ScreenName } from "../const";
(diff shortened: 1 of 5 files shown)


juan-cortes (approved):

ychen-ledger (commented):
Validated with:
- Version 2.29.1 (develop mode)
- Android version: 10
- FW 1.3.0
- Ethereum

ychen-ledger (approved):

PR #1803, opened 2021-07-06
BUG: Fixing catalogue name from 'Browse' to 'Discover'

lruggiero-ledger:
### Type
Fixing catalogue name from 'Browse' to 'Discover'

### Context
Bug Fix

No test plan changes. Just naming on bottom menu

### DIFF
--- a/src/locales/ja/common.json
+++ b/src/locales/ja/common.json
@@ -1213,7 +1213,7 @@
     "transfer": "Transfer",
     "manager": "Manager",
     "settings": "Settings",
-    "platform": "Browse",
+    "platform": "Discover",
     "nanoX": "Nano X"
   },
   "portfolio": {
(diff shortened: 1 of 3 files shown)


src/locales/ja/common.json
```diff
@@ -1213,7 +1213,7 @@
     "transfer": "Transfer",
     "manager": "Manager",
     "settings": "Settings",
-    "platform": "Browse",
+    "platform": "Discover",
```
gre, line 1216: changes not to be done on the locales* other than en, otherwise crowdin would consider it as "translated".

thanks.

gre: (these file are auto generated and crowdin will send automatically a PR to make them changed to "discover" too)

gre (changes requested):

PR #1827, opened 2021-07-16
LL-6423 - Add Platform DApp disclaimer

klambert-ledger:
We should not be able to open a DApp without consenting to the disclaimer. Disclaimer can be accepted once or dismissed forever for all DApps.

### Videos
<details>
  <summary>iOS 📸 </summary>

https://user-images.githubusercontent.com/86959861/125981848-6d04ca2e-e49e-47b6-a25f-ee51a1887f3c.mp4
</details>
<details>
  <summary>Android 📸 </summary>

https://user-images.githubusercontent.com/86959861/125981882-48e9652b-3e86-41d6-8d7f-0038cae69cf0.mp4

</details>

### Type

Feature

### Context

[LL-6423](https://ledgerhq.atlassian.net/browse/LL-6423)

Platform -> Press on any DApp

### DIFF
--- a/src/screens/Platform/DAppDisclaimer.js
+++ b/src/screens/Platform/DAppDisclaimer.js
@@ -1,0 +1,222 @@
+// @flow
+
+import React, { useCallback, useState } from "react";
+
+import { useTheme } from "@react-navigation/native";
+import { Trans, useTranslation } from "react-i18next";
+import { StyleSheet, View, TouchableOpacity } from "react-native";
+
+import BottomModal from "../../components/BottomModal";
+import Checkbox from "../../components/CheckBox";
+import Button from "../../components/Button";
+import LiveIcon from "../../icons/LiveLogo";
+import LText from "../../components/LText";
+import CloseIcon from "../../icons/Close";
+import InfoIcon from "../../icons/Info";
+import { rgba } from "../../colors";
+import AppIcon from "./AppIcon";
+
+export type Props = {
+  closeDisclaimer: () => void,
+  disableDisclaimer: () => void,
+  onContinue: () => void,
+  isOpened: boolean,
+  icon?: string | null,
+};
+
+const DAppDisclaimer = ({
+  closeDisclaimer,
+  isOpened,
+  disableDisclaimer,
+  onContinue: next,
+  icon,
+}: Props) => {
+  const { t } = useTranslation();
+  const { colors } = useTheme();
+  const [
+    disableDisclaimerChecked,
+    setDisableDisclaimerChecked,
+  ] = useState<boolean>(false);
+
+  const onClose = useCallback(() => {
+    closeDisclaimer();
+  }, [closeDisclaimer]);
+  const onContinue = useCallback(() => {
+    if (disableDisclaimerChecked) {
+      disableDisclaimer();
+    }
+    closeDisclaimer();
+    next();
+  }, [disableDisclaimerChecked, closeDisclaimer, disableDisclaimer, next]);
+
+  return (
+    <BottomModal
+      style={{
+        ...styles.root,
+        backgroundColor: colors.card,
+      }}
+      isOpened={isOpened}
+      onClose={onClose}
+    >
+      <TouchableOpacity style={styles.closeIcon} onPress={onClose}>
+        <CloseIcon size={18} />
+      </TouchableOpacity>
+
+      <View style={[styles.flowRow, styles.head]}>
+        <View style={[styles.headElements, styles.roundLogoContainer]}>
+          <LiveIcon size={31} color="blue" />
+        </View>
+        {icon && (
+          <>
+            <View style={[styles.headElements, styles.flowRow]}>
+              {Array(6)
+                .fill()
+                .map((_, i) => (
+                  <View
+                    key={i}
+                    style={{ ...styles.dash, backgroundColor: colors.live }}
+                  />
+                ))}
+            </View>
+            <View style={styles.headElements}>
+              <AppIcon size={48} icon={icon} />
+            </View>
+          </>
+        )}
+      </View>
+
+      <LText semiBold style={[styles.textCenter, styles.title]}>
+        <Trans i18nKey={"platform.disclaimer.title"} />
+      </LText>
+      <LText style={[styles.textCenter, styles.description]}>
+        <Trans i18nKey={"platform.disclaimer.description"} />
+      </LText>
+
+      <View
+        style={{
+          ...styles.flowRow,
+          ...styles.blueInfo,
+          backgroundColor: rgba(colors.live, 0.1),
+        }}
+      >
+        <View style={styles.blueInfoIcon}>
+          <InfoIcon color={colors.live} size={18} />
+        </View>
+        <LText style={{ ...styles.blueInfoText, color: colors.live }}>
+          <Trans i18nKey={"platform.disclaimer.legalAdvice"} />
+        </LText>
+      </View>
+
+      <TouchableOpacity
+        style={[styles.flowRow, styles.reminder]}
+        onPress={() => setDisableDisclaimerChecked(!disableDisclaimerChecked)}
+      >
+        <Checkbox
+          style={styles.checkbox}
+          iconCheckSize={12}
+          isChecked={disableDisclaimerChecked}
+        />
+        <LText semibold style={styles.reminderText}>
+          <Trans i18nKey={"platform.disclaimer.checkbox"} />
+        </LText>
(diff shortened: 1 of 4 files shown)


src/screens/Platform/DAppDisclaimer.js
```diff
@@ -0,0 +1,223 @@
...
+        <LText style={{ ...styles.blueInfoText, color: colors.live }}>
+          <Trans i18nKey={"platform.disclaimer.legalAdvice"} />
+        </LText>
+      </View>
+
+      <TouchableOpacity
+        style={[styles.flowRow, styles.reminder]}
+        onPress={() => {
+          setDisableDisclaimerChecked(!disableDisclaimerChecked);
+        }}
```
achabrolin-ledger, line 113: Since the new state is computed from the previous state, you could use a [functional update](https://reactjs.org/docs/hooks-reference.html#functional-updates).

```suggestion
        onPress={() => setDisableDisclaimerChecked(state => !state)}
```


Or at least this could be written as a one liner (without curly braces)
```suggestion
        onPress={() => setDisableDisclaimerChecked(!disableDisclaimerChecked)}
```

src/screens/Platform/DAppDisclaimer.js
```diff
@@ -0,0 +1,223 @@
...
+            <View style={[styles.headElements, styles.flowRow]}>
+              {Array(6)
+                .fill()
+                .map((_, i) => (
+                  <View
+                    key={i}
+                    style={{ ...styles.dash, backgroundColor: colors.live }}
+                  />
+                ))}
+            </View>
```
achabrolin-ledger, line 79: You should memoize this "dashs" component to avoid this array.map on each render.

```suggestion
            <View style={[styles.headElements, styles.flowRow]}>{Dashs}</View>
```

Here is one way to do so:
```js
  const Dashs = useMemo(
    () =>
      [...Array(6)].map((_, i) => (
        <View
          key={i}
          style={{ ...styles.dash, backgroundColor: colors.live }}
        />
      )),
    [colors.live],
  );
```

achabrolin-ledger (changes requested):
We might have some re-render / performance issue here. The `PlatformCatalog` component (located here: `src/screens/Platform/index.js`) is re-rendered every time we open / close the disclaimer modal. Since this screen contains a list of all available apps (which will only grow in the future), it would be wise to address this.

It might be related to the updates done to `handlePressCard` which update the state of the component every time, thus triggering a re-render of the whole screen.

Bellow are two recordings illustrating this issue


<details>
<summary>Before</summary>

https://user-images.githubusercontent.com/87011321/126497401-ece2b408-3dbb-40da-bebd-60d3b9a4cd5f.mov

</details>

<details>
<summary>After</summary>

https://user-images.githubusercontent.com/87011321/126497708-f1306d9a-dccf-4894-b37c-14b3981d9039.mov

</details>

src/screens/Platform/DAppDisclaimer.js
```diff
@@ -0,0 +1,223 @@
...
+    next();
+  }, [disableDisclaimerChecked, closeDisclaimer, disableDisclaimer, next]);
+
+  return (
+    <BottomModal
+      style={{
+        ...styles.root,
+        backgroundColor: colors.card,
+      }}
+      isOpened={isOpened}
```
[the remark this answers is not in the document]
klambert-ledger: Yep forgot it like on the info panel. Will add it 👍

src/screens/Platform/DAppDisclaimer.js
klambert-ledger: Using the state feels way less intuitive since we already declared the boolean before. I'll change for the one liner 👍

src/screens/Platform/DAppDisclaimer.js
klambert-ledger: IMO the cost of memoization is way worse than 2 consecutive loops on 6 indexes. Memoization should be done either on state related data in order to work with the hooks api or for performance consuming tasks, and to me this is none of those case.

If we really want to avoid these loops, I can always declare the array outside of the component, but it feels pretty nitpicky to me 😜

klambert-ledger:
> We might have some re-render / performance issue here. The `PlatformCatalog` component (located here: `src/screens/Platform/index.js`) is re-rendered every time we open / close the disclaimer modal. Since this screen contains a list of all available apps (which will only grow in the future), it would be wise to address this.
> 
> It might be related to the updates done to `handlePressCard` which update the state of the component every time, thus triggering a re-render of the whole screen.
> 
> Bellow are two recordings illustrating this issue
> 
> Before
>  before.mov 
> After

You're right, I'm gonna investigate this. 👀

src/screens/Platform/DAppDisclaimer.js
achabrolin-ledger: It looks like it could still be optimised because as of today, this whole logic (computation + render) is done every time an action happens.

cf. 
<details>
<summary>Before memoization</summary>

https://user-images.githubusercontent.com/87011321/126509064-d9ba7579-ee00-484a-a1a7-fb78aa86ac66.mov

</details>

<details>
<summary>After memoization</summary>

https://user-images.githubusercontent.com/87011321/126509043-131ed775-0f82-4973-ac69-fb356fc04ae9.mov

</details>

It would be interesting to run the Profiler to test any performance drop or optimisation related to this.

klambert-ledger: Yeah that's not really what I meant by the cost of memoization. I was talking about this kind of point: https://kentcdodds.com/blog/usememo-and-usecallback.
Sure we are re-rendering 6 empty "divs" pretty much, but the cost of "optimizing" this with a memoization will probably (surely) have no benefit on performances and might actually make it worse from what I can tell.

I will memoize it to make your logs disappear but I believe it won't have the effect you think it has 😆

klambert-ledger:
> > We might have some re-render / performance issue here. The `PlatformCatalog` component (located here: `src/screens/Platform/index.js`) is re-rendered every time we open / close the disclaimer modal. Since this screen contains a list of all available apps (which will only grow in the future), it would be wise to address this.
> > It might be related to the updates done to `handlePressCard` which update the state of the component every time, thus triggering a re-render of the whole screen.
> > Bellow are two recordings illustrating this issue
> > Before
> > before.mov
> > After
> 
> You're right, I'm gonna investigate this. 👀

@achabrolin-ledger After some investigation, there is no unnecessary re-render or performance issue. The hooks API is doing its job by triggering a recomputation on state change, but there is no re-render here. 

To show the difference here is a little example where every rendering will paint :


https://user-images.githubusercontent.com/86959861/126524294-dd4afa16-c0d8-4a8f-8c28-3d4d81f7b4ff.mov


(you can mess around with if you want to test here: https://bit.ly/36XYUgZ)

We could if you want Memoize the AppCards at the first rendering but as said before, it feels like pre-optimization to me and I'm really not sure that it would have any benefit.

achabrolin-ledger:
> > > We might have some re-render / performance issue here. The `PlatformCatalog` component (located here: `src/screens/Platform/index.js`) is re-rendered every time we open / close the disclaimer modal. Since this screen contains a list of all available apps (which will only grow in the future), it would be wise to address this.
> > > It might be related to the updates done to `handlePressCard` which update the state of the component every time, thus triggering a re-render of the whole screen.
> > > Bellow are two recordings illustrating this issue
> > > Before
> > > before.mov
> > > After
> > 
> > 
> > You're right, I'm gonna investigate this. 👀
> 
> @achabrolin-ledger After some investigation, there is no unnecessary re-render or performance issue. The hooks API is doing its job by triggering a recomputation on state change, but there is no re-render here because there is no change for the renderer when diffing the returned value of the component.
> 
> To show the difference here is a little example where every rendering will paint :
> 
>  Screen.Recording.2021-07-21.at.18.22.28.mov 
> (you can mess around with if you want to test here: https://bit.ly/36XYUgZ)
> 
> We could if you want Memoize the AppCards at the first rendering but as said before, it feels like pre-optimization to me and I'm really not sure that it would have any benefit.
> 
> EDIT: Oh and while I'm thinking about it, the reason why you're not seing logs appearing on your "before" video is because of how React Native is working with its navigation. This is called a `stack` which means that the view is still under the DApp browser when it's opening. So opening or closing the DApp browser will not change the Catalog view at any point.

@klambert-ledger Thanks for the investigation and the exemple! According to this exemple it indeed looks like there is no repaint, contrary to my belief. If there is no significant performance issue then we can keep it as is 👌

achabrolin-ledger (approved):

PR #1829, opened 2021-07-19
feat(manager): use displayName in applist instead of name

valpinkman:
use display_name in appList instead of name

needs https://github.com/LedgerHQ/ledger-live-common/pull/1309

### Type

Improvement

### Context

[LL-6299]

Manager => AppList

[LL-6299]: https://ledgerhq.atlassian.net/browse/LL-6299


gre:
included in #1832

PR #1855, opened 2021-08-09
LL-6534 Translations bugs

juan-cortes:
Unified pr to fix translation bugs reported, please keep your commits atomic or create sub prs to this branch if the task is more involved. If QA checks any of these and it's not clear enough what to test, please comment for clarification.

- [**LL-6555**](https://ledgerhq.atlassian.net/browse/LL-6555) Remove modal about the language warning for non-ENG
    - [x] Dev
    - [x] QA
- [**LL-6557**](https://ledgerhq.atlassian.net/browse/LL-6557) Display language name instead of locale in settings
    - [x] Dev
    - [x] QA
- [**LL-6588**](https://ledgerhq.atlassian.net/browse/LL-6588) Remove address count from "From/To" on op details 
    - [x] Dev
    - [x] QA
- [**LL-6553**](https://ledgerhq.atlassian.net/browse/LL-6553) Theme settings options and Fees strategies not translated + Translation not happening when changing language in settings (cf. videos bellow) 
    - [x] Dev
    - [x] QA
- ~~[LL-6574](https://ledgerhq.atlassian.net/browse/LL-6574) LLD - dates should be translatable~~ Not getting done in this PR as it requires further work
  - ~~[x] Dev~~
  - ~~[ ] QA~~

---

**Translation not happening when changing language in settings**

<details>
  <summary>Before 🎥 </summary>

https://user-images.githubusercontent.com/87011321/128722225-1e90d655-281e-4305-8816-ca371aabce12.mov

</details>

<details>
  <summary>After 🎥 </summary>

https://user-images.githubusercontent.com/87011321/128722321-90202f13-fed3-411c-a260-9fca6ac80708.mov
  

</details>

---

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-6534

Once we have them all or a bunch of them we can ship a build to the translation team for validation.


achabrolin-ledger (dismissed):

src/locales/en/common.json
```diff
@@ -1327,8 +1327,8 @@
     "failed": "Failed",
     "fees": "Network fees",
     "noFees": "No fee",
-    "from": "From ({{count}})",
-    "to": "To ({{count}})",
```
ggilchrist-ledger, line 1331: @juan-cortes These are still visible on non-English languages

juan-cortes: Indeed, we update the English one and it will be propagated through crowdin translations, we don't handle the translations manually.

gre: Only the en file is the source indeed. Other will be updated by crowdin.

valpinkman (approved):

src/utils/hooks/useDateTimeFormat.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import { useSelector } from "react-redux";
+import { languageSelector } from "../../reducers/settings.js";
+
+const useDateTimeFormat = (options?: Intl$DateTimeFormatOptions = {}) => {
+  const currentLanguage = useSelector(languageSelector);
+  const formatter = new Intl.DateTimeFormat(currentLanguage, options);
+
+  return formatter.format;
+};
```
gre, line 10: we should always use a cache. date formatting have an history of being a bottleneck in our app and this is why was had a component that is memo'ed in the past.

gre (changes requested):

src/utils/hooks/useDateTimeFormat.js
gre: moreover, it might be a blocker to do Intl if we want to go Hermes , https://github.com/facebook/hermes/issues/23

jgiamporcaro-ledger: Thanks for your comments!
In fact, after reading your responses from Slack I double-checked the Android supports and discovered that we have an issue with the configuration, `Intl` isn't injected as expected by our configuration. I have the feeling that the configuration which is currently located in `build.grade` files is outdated for this specific use case and should be updated as described [here](https://github.com/react-native-community/jsc-android-buildscripts#international-variant).

I recommend removing the commit related to `Intl` from this commit and open a new dedicated PR as the scope is bigger than expected.

@gre Regarding Android configuration from `build.gradle` files, is it the job of the core team to update it or can I put my hands in? I can see the configuration to run Intl from RNv0.59.x already in place, I think that we just have to bump it to supports version 0.60.x+.

jgiamporcaro-ledger: Here is the new [PR](https://github.com/LedgerHQ/ledger-live-mobile/pull/1858)

ggilchrist-ledger:
LGTM

valpinkman (approved):

PR #1837, opened 2021-07-22
live-common with typescript

valpinkman:
First iteration of live common typescript rework

PR #1858, opened 2021-08-10
LL-6574 - Dates should be translatable - Intl setup improvement

jgiamporcaro-ledger:
Scope: https://github.com/LedgerHQ/ledger-live-mobile/pull/1855#discussion_r686072322
### Type

Feature/Bug Fix

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -123,7 +123,7 @@ def enableProguardInReleaseBuilds = false
  * give correct results when using with locales other than en-US.  Note that
  * this variant is about 6MiB larger per architecture than default.
  */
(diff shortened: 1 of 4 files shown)


gre:
i think we'll have to postpone this part for now.
- i see it will collide with #1837 
- if we want to move to hermes this would have to change again i think

i see the votes was "1" & was probably underrated too. cc @LedgerHQ/live-squad-yellow

jgiamporcaro-ledger:
As decided during the yellow sprint meeting this task is postponed.

jgiamporcaro-ledger:
Hardly postponed.

PR #1864, opened 2021-08-13
elrond as supported currency

AlexandruPislariu:


StanislavSava:
@valpinkman It seems that the e2e test and bundle app fails, although we didn't touch neither the Podfile.lock file, nor the yarn.lock/package.json files. Do you have an idea on what might be the issue? I also get the same error if I try to "yarn" the app on a fresh clone from the develop branch of this repo.
Screenshot 2021-09-08 at 16 10 14

haammar-ledger:
Replaced with https://github.com/LedgerHQ/ledger-live-mobile/pull/1895

PR #1865, opened 2021-08-13
don't reportError if it's not instance of error

gre:
see https://github.com/getsentry/sentry-javascript/issues/2292


gre:
there is no qa needed on this. we simply have a sentry issue here of reporting non Error objects.

PR #1873, opened 2021-08-23
LL-6669 rebrand loader

nparigi-ledger:
https://user-images.githubusercontent.com/89014981/130453062-40c3c0ee-bbbe-4bae-86a1-4a38f8b399df.mp4

### Type

Rebranding component

### Context

https://ledgerhq.atlassian.net/browse/LL-6669

uilib / storybook


uilib/src/components/Loader/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+const ProgressLoader = ({
+  progress = 0,
+  onPress,
+  displayCancelIcon = true,
+}: Props): React.ReactElement => {
+  const { colors } = useTheme();
+  const backgroundColor = colors.palette.primary.backgroundLight;
+  const progressColor = colors.palette.primary.dark;
+
+  const strokeWidth = 4;
```
juan-cortes, line 28: This doesn't seem to be matching the Figma, can you double-check the dimensions?
Also check with design in case we want to support multiple sizes 🤔 

image

nparigi-ledger: You are right, the stroke width is 2, not 4, thanks :pray: 
For the custom size, i assumed that this loader button would be used only in https://www.figma.com/file/wGzwuUVo0rkCJ3sM8cpbDY/Library---LLM?node-id=672%3A552 but it could be useful to support this indeed, i will check this.

uilib/src/components/Loader/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const normalizedRadius = R - strokeWidth / 2;
+  const circumference = normalizedRadius * 2 * Math.PI;
+  const strokeDashoffset = circumference - progress * circumference;
+  return (
+    <TouchableOpacity onPress={onPress}>
```
jelbaz-ledger, line 33: Wondering if the loader component should only be touchable if it has an `onPress` prop (and make the prop optional btw) 🤔. I think most of the time a loader is not meant to be interactive?

uilib/src/components/Loader/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  progress?: number;
+
+  // function triggered when pressing the loader
+  onPress: () => void;
+
+  // Display the square in the middle of the loader
+  displayCancelIcon?: boolean;
+};
+
+const R = 25;
```
achabrolin-ledger, line 17: What does `R` stands for here? Could you provide a more precise / specific name?

If it represents a radius, a better name could be `RADIUS`

achabrolin-ledger (changes requested):

uilib/src/components/Loader/index.tsx
nparigi-ledger: Fixed, thanks :+1:

uilib/src/components/Loader/index.tsx
nparigi-ledger: Part of the old loader code (progressCircle), it's fixed, thanks

uilib/src/components/Loader/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  onPress,
+  displayCancelIcon = true,
+}: Props): React.ReactElement => {
+  const { colors } = useTheme();
+  const backgroundColor = colors.palette.primary.backgroundLight;
+  const progressColor = colors.palette.primary.dark;
+
+  const strokeDashoffset = circumference - progress * circumference;
+  return (
+    <TouchableOpacity disabled={!onPress} {...{onPress} }>
```
valpinkman, line 33: why this syntax ?
```jsx
  <TouchableOpacity disabled={!onPress} {...{onPress} }>
```

this is more readable i believe
```suggestion
    <TouchableOpacity disabled={!onPress} onPress={onPress}>
```

valpinkman (changes requested):

uilib/src/components/Loader/index.tsx
nparigi-ledger: Indeed, i was afraid of crashing if onPress was null, but if it's disabled already ... my mind was elsewhere, fixed :+1:

valpinkman (approved):

PR #1877, opened 2021-08-26
ENGB2C-104 Swap form

LFBarreto:
- LL-6623 integrate swap form v2 tab navigation

### Context

[ENGB2C-104]

[LL-4684]: https://ledgerhq.atlassian.net/browse/LL-4684

[LL-6623]: https://ledgerhq.atlassian.net/browse/LL-6623

[LL-6624]: https://ledgerhq.atlassian.net/browse/LL-6624
[LL-6632]: https://ledgerhq.atlassian.net/browse/LL-6632

[LL-6630]: https://ledgerhq.atlassian.net/browse/LL-6630


gre (dismissed):
unclear why blockchainsupport team requested by ✅  from me.

src/screens/Swap/FormSelection/SelectAccountScreen.js
```diff
@@ -0,0 +1,253 @@
...
+  const { swap, target, selectedCurrency, setAccount, provider } = route.params;
+
+  const unfilteredAccounts = useSelector(accountsSelector);
+  const selectableCurrencies = useSelector(swapSelectableCurrenciesSelector);
+
+  const accounts = unfilteredAccounts.filter(
+    acc => acc && selectableCurrencies.includes(getAccountCurrency(acc).id),
+  );
+
+  const enhancedAccounts = useMemo(() => {
```
elbywan, line 53: `useMemo` seems useless here (unless I'm mistaken) since the `accounts` dependency reference will change every time the component renders

src/screens/Swap/OperationDetails.js
```diff
@@ -62,8 +62,8 @@ const OperationDetails = ({ route }: Props) => {
   const swap = fromAccount.swapHistory.find(s => s.swapId === swapId);
   const status = Config.DEBUG_SWAP_STATUS || swap.status;
 
-  const fromCurrency = getAccountCurrency(fromAccount);
-  const toCurrency = getAccountCurrency(toAccount);
+  const fromCurrency = fromAccount && getAccountCurrency(fromAccount);
```
elbywan, line 65: Is the new condition because of flow? I see `fromCurrency.type` just below (line 71 - `const url = …`) that would crash if fromCurrency is undefined anyway I think?

elbywan (dismissed):
Only minor stuff 👍

src/screens/Swap/FormSelection/SelectAccountScreen.js
LFBarreto: i'll memo the accounts as well then :grimacing:

src/screens/Swap/OperationDetails.js
LFBarreto: good catch i'll check the condition everywhere :+1: 
during testing i went in a case where fromAccount was null during a frame for a reason i dont know.
i prefer be sure than sorry here

elbywan (approved):
👌

PR #1880, opened 2021-08-30
LL 6474 - Add a banner alert on portfolio that points to LLD to update the firmware

klambert-ledger:
LLM will now have a top banner to try to incent people to update their ledger through LLD.

### Type

UI Polish

### Context

[LL-6474](https://ledgerhq.atlassian.net/browse/LL-6474)

1. Open the manager with an outdated Ledger.
2. The banner should appear and stay on every page. The banner should be pressable to open a drawer and closable.
3.  Close the app, and open it again.
4. The banner should appear again.
5.  (Optional) The banner should disappear when using an updated device again (So either go to 1. with a new device or the same updated one)


src/components/FirmwareUpdateBanner.js
```diff
@@ -0,0 +1,282 @@
...
+
+  return (
+    showBanner && (
+      <>
+        <Animated.View
+          style={[
+            styles.banner.root,
+            {
+              opacity,
+              transform: [{ translateY }],
```
LFBarreto, line 105: if its just a simple slide down you could've used [react-native-animatable](https://github.com/oblador/react-native-animatable)
which we already use in the app :sweat_smile: 
If this works fine keep it and happy to see you using reanimated but in case of simple animations keep in mind there are simpler alternatives.

LFBarreto (dismissed):

src/components/FirmwareUpdateBanner.js
klambert-ledger: Sure ! I just used the same tools that were used by the SnackBar.js, but it felt quite a lot of boilerplate just for this so I'd be happy to switch it if there is something simpler 👍

klambert-ledger: Switched to animatable ! 👍 
Thx for the input !

valpinkman (changes requested):
Ready to merge when CI passes

src/locales/en/common.json
```diff
@@ -2747,6 +2747,11 @@
       "flash-bootloader": "Bootloader updating",
       "flash": "Flashing your device",
       "firmware": "Firmware updating"
+    },
+    "newVersion": "Update Firmware to {{version}} is available",
+    "drawerUpdate": {
+      "title": "Firmaware Update",
+      "description": "Update your Ledger Nano firmware by connecting it to Ledger Live application on desktop"
```
ggilchrist-ledger, line 2754: Just a small gramatical thing: could you change it to `"Update your Ledger Nano firmware by connecting it to _the_ Ledger Live application on desktop"` @klambert-ledger

lambertkevin: Done 👍

src/locales/en/common.json
```diff
@@ -2747,6 +2747,11 @@
       "flash-bootloader": "Bootloader updating",
       "flash": "Flashing your device",
       "firmware": "Firmware updating"
+    },
+    "newVersion": "Update Firmware to {{version}} is available",
+    "drawerUpdate": {
+      "title": "Firmaware Update",
```
ggilchrist-ledger, line 2753: Just noticed it says Firmaware too :D Could you fix that? @klambert-ledger

lambertkevin: Done 👍

ggilchrist-ledger:
When I do the onboarding journey with an out of date Nano, the banner doesn't appear straight away. It only appears once I've accessed it from the manager. Is this expected? It might just be that a device is only counted as `lastSeen` when it has been opened in the Manager, however I would expect the banner to be visible if the out of date Nano was used to setup the app. @lambertkevin

ggilchrist-ledger:
LGTM (card created here to track potential UX improvements - https://ledgerhq.atlassian.net/secure/RapidBoard.jspa?rapidView=277&projectKey=LL&view=planning&selectedIssue=LL-7074&versions=visible&issueLimit=100)

ggilchrist-ledger (approved):

PR #1906, opened 2021-09-16
LL 6631 swap form - select fees screen

nparigi-ledger:
https://user-images.githubusercontent.com/89014981/133661899-47a2bec0-a37c-4b16-aa72-0b5f8f513c5b.mp4

### Context

https://ledgerhq.atlassian.net/browse/LL-6631


LFBarreto (changes requested):

src/components/SelectFeesStrategy.js
```diff
@@ -89,19 +84,35 @@ export default function SelectFeesStrategy({
         styles.feeButton,
         {
           borderColor:
-            feesStrategy === item.label ? colors.live : colors.background,
-          backgroundColor:
-            feesStrategy === item.label ? colors.lightLive : colors.lightFog,
+            feesStrategy === item.label
+              ? colors.live
+              : rgba(colors.contrastBackground, 0.1),
```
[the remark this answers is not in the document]
nparigi-ledger: Done

src/components/SelectFeesStrategy.js
```diff
@@ -89,19 +84,35 @@ export default function SelectFeesStrategy({
...
+            name={"tachometer-alt-fastest"}
           />
-          <LText semiBold style={styles.feeLabel}>
+          <LText
+            bold
+            style={[
+              styles.feeLabel,
+              {
+                color:
+                  feesStrategy === item.label
```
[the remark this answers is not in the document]
nparigi-ledger: I reverted the file (:cry:)

src/components/RootNavigator/BaseNavigator.js
```diff
@@ -156,6 +157,16 @@ export default function BaseNavigator() {
           headerRight: null,
         }}
       />
+      <Stack.Screen
+        name={ScreenName.SwapV2FormSelectFees}
+        component={SwapFormV2SelectFees}
+        options={{
+          headerTitle: () => (
+            <StepHeader title={t("transfer.swap.form.summary.fees")} />
```
juan-cortes, line 165: I haven't followed the rework of mobile as much as desktop but we need to be completely detached from the components and wording of the previous version, and be able to toggle between them. Wording should be namespaced to v2 like on lld

src/locales/en/common.json
```diff
@@ -1923,7 +1923,7 @@
           "receiveFloat": "Amount to receive before service fees",
           "provider": "Provider",
           "method": "Rate",
-          "fees": "Max fees",
+          "fees": "Fees",
```
juan-cortes, line 1926: This is exactly what I meant, by changing wording here, you are altering the v1 text too and that shouldn't happen at all. Please sync with @LFBarreto in case he wants to do things differently but to me we need a swapv2 namespace in the wording file that has nothing to do with the v1

src/screens/Swap2/FormSelection/SelectFeesScreen.js
```diff
@@ -0,0 +1,76 @@
...
+  allAccounts: AccountLikeArray,
+  navigation: any,
+  route: { params: SwapRouteParams },
+};
+
+export default function SelectFees({ navigation, route }: Props) {
+  const { transaction } = route.params;
+  const { account, parentAccount } = route.params;
+
+  // Tmp: fixed networkInfo while waiting for ll-common to build complete transaction info
```
juan-cortes, line 29:
```suggestion
  // FIXME : Tmp fixed networkInfo while waiting for ll-common to build complete transaction info
```

src/screens/Swap2/index.js
```diff
@@ -268,6 +267,18 @@ export default function SwapForm({
     });
   }, [exchange, navigation]);
 
+  const onEditFees = () => {
+    console.log("send transaction", transaction);
```
juan-cortes, line 271: Remove the console log

juan-cortes (changes requested):
Although it might be outside of the scope of this pr, wording changes should not impact v1 of the swap and everything should be atomic. There seems to be lint issues too in the navigator file, please do a check to make sure ci passes

src/locales/en/common.json
nparigi-ledger: I changed it, as it was not used anywhere else.
But you are right, it's not a good thing, if we need to keep the 2 swap in the codebase, we should have a different namespace.
I will create a swapv2 namespace

src/screens/Swap2/FormSelection/SelectFeesScreen.js
nparigi-ledger: Removing the whole Fixme code, it was a mistake on my side

src/screens/Swap2/index.js
nparigi-ledger: :ok_hand:

juan-cortes (approved):

PR #1922, opened 2021-09-28
LL-4045 Automatically display keyboard on password screens

juan-cortes:
> The branch number is wrong because I'm a moron.

The current behavior of the keyboard when entering the password screens is not ideal. The autofocus only takes effect on the first screen and will not show the keyboard on the confirmation one. I refactored the component (without judging the logic) into a functional component and will trigger the focus manually when mounting. This seems to solve the issue.

### Type

Bug Fix

### Context
https://ledgerhq.atlassian.net/browse/LL-4045

https://user-images.githubusercontent.com/4631227/135092916-9db9bd0d-f3c7-441b-86bc-9154f48f2738.mp4

- On an instance of LLM without a password setup a password
- Confirm that both screens (password + password confirmation) start with the keyboard open like in the video
- Confirm that we can still set up a password and remove it.
- Test both iOS (have not tested iOS) and Android

### DIFF
--- a/src/components/PasswordInput.js
+++ b/src/components/PasswordInput.js
@@ -1,5 +1,5 @@
 /* @flow */
-import React, { PureComponent } from "react";
+import React, { useEffect, useState, useCallback, useRef } from "react";
 import { View, StyleSheet, TextInput } from "react-native";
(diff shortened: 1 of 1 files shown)


LFBarreto (approved):
:guitar:

PR #1976, opened 2021-11-10
LL-6956 Fix first pairing failure bug

juan-cortes:
We've had this bug forever and it impacts 100% of our new pairings, impacting the UX of all our new users negatively. Just returning the full device instead of the deviceId allows the subsequent device action to actually connect when entering the manager and solves the issue. Just had to add a bunch of traces to see where it was breaking.

### Type

Bug fix 🐛 

### Context

[](https://ledgerhq.atlassian.net/browse/LL-6956)

### DIFF
--- a/src/screens/PairDevices/index.js
+++ b/src/screens/PairDevices/index.js
@@ -43,7 +43,7 @@ type PairDevicesProps = {
 };
 
 type RouteParams = {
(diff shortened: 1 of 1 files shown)


achabrolin-ledger (approved):
🚀

gre (approved):

PR #1982, opened 2021-11-16
LL-8135 Fix crash after import

juan-cortes:
Due to a missed step in the [migration guide from v5.x to v6.x of `react-navigation` ](https://reactnavigation.org/docs/upgrading-from-5.x/#some-exports-are-now-moved-to-the-element-library) we were using an incorrect import path for the `HeaderBackButton` component, meaning the app would try to wrap undefined and crash whenever we relied on that import.

image

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-8135

- Import accounts via live qr
- Make sure the app doesn't crash on the result screen.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "@react-native-community/masked-view": "^0.1.11",
     "@react-native-community/netinfo": "^6.0.1",
     "@react-navigation/bottom-tabs": "^6.0.9",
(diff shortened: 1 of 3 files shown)


LFBarreto (approved):
:rocket:

PR #2020, opened 2021-11-25
LL-8124 (mobile) Allow to add 1 empty account for each derivation mode

ghost:
https://ledgerhq.atlassian.net/browse/LL-8124

:information_source: Non-breaking with live-common, but requires https://github.com/LedgerHQ/ledger-live-common/pull/1539 for the fix to work.

Allow the user to add 1 empty account for each derivation mode (address type) of the currency, even if an empty account is already added.

#### Preferred type displayed by default:

Screenshot_1639136272

#### All available empty types:

Screenshot_1637855060


gre (approved):

juan-cortes:
Didn't we move to only showing an empty account of a specific derivation mode a few months ago?

haammar-ledger:
@juan-cortes Not sure if that's what you mean, but before displaying all available empty accounts we still only show the first preferred type, and there's a hidden section to open to display other types. I didn't change this. I've updated the screenshots, see description, does it answer your question?

LFBarreto (approved):

PR #2029, opened 2021-11-30
Remove unused dependencies

ghost:
Just remove `@ledgerhq/compressjs` and `@ledgerhq/cryptoassets` from dependencies since they're not directly used.

### Type

Code cleanup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,6 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


gre:
good idea. thanks. i've included it in https://github.com/LedgerHQ/ledger-live-mobile/pull/2038.

PR #2006, opened 2021-11-23
LL-7746 - Rebranding - Onboarding welcome and language picker

jelbaz-ledger:
### Rebranding: onboarding welcome + onboarding language picker

**➕ Also:**
- made a reusable bottom modal using react navigation
- updated react native reanimated to match the native-ui version (prevents duplicate packages)
- upgraded pods
- linked the fonts

----- 

#### Movies 🎥 

- **IOS**

https://user-images.githubusercontent.com/86958797/143074716-e9eef76a-19e9-46e0-8df4-f3f5f9a4a1d6.mp4

- **Android**

### Type

UI

### Context

[LL-7746]
[Figma](https://www.figma.com/file/x9nwU0JtDtyUIhAyBB2El8/LLM-%2F-Onboarding?node-id=3%3A9)

### Parts of the app affected / Test plan

First 2 onboarding screens

[LL-7746]: https://ledgerhq.atlassian.net/browse/LL-7746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2035, opened 2021-12-02
LL-7746 - Rebrand - Improve onboarding modals animations

jelbaz-ledger:
#### As a follow up to #2006 

- improves navigation experience (pressing the previous button from the terms now goes back to the first welcome screen - not the modal)
- improves transitions - only the content is transitioned (not the header) and uses native animation style
- overall improvements to the modal container & header components

https://user-images.githubusercontent.com/86958797/144416158-3edc172d-3bcf-48d9-9822-45e567d1cb47.mp4

### Type

UI

### Context

[LL-7746]

[LL-7746]: https://ledgerhq.atlassian.net/browse/LL-7746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/NavigationHeader.tsx
+++ b/src/components/NavigationHeader.tsx
@@ -1,17 +1,13 @@
 import React from "react";
 import { useTranslation } from "react-i18next";
-import { Pressable, View } from "react-native";
+import { View } from "react-native";
 import { ArrowLeftMedium, CloseMedium } from "@ledgerhq/native-ui/assets/icons";
 import { Flex, Text, Link } from "@ledgerhq/native-ui";
 import { StackHeaderProps } from "@react-navigation/stack";
 import { getHeaderTitle } from "@react-navigation/elements";
 
-// Magic number
-const HEADER_HEIGHT = 52;
-
 function NavigationHeader({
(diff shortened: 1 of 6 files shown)


src/components/NavigationHeader.tsx
```diff
@@ -20,36 +16,31 @@ function NavigationHeader({
...
-          {title}
-        </Text>
-        <Link size="large" Icon={CloseMedium} onPress={navigation.popToTop} />
-      </Flex>
     </Flex>
   );
 }
 
-export default NavigationHeader;
+export default (props: StackHeaderProps) => <NavigationHeader {...props} />;
```
qd-qd, line 46: Why not just return the component? 🤔

qd-qd (approved):

src/components/NavigationHeader.tsx
elbywan: Because it causes hook issues due to a poor react navigation implementation. :/

PR #2062, opened 2021-12-16
Add documentation about env variables & working on iOS/Android emulators

OlivierFreyssinet:
- Add link to a more exhaustive documented list of environment variables (in live common)
- Remove dead link for `DEVICE_PROXY_URL` env variable explanation
- Add explanations on how to debug on an emulator using ledger-live cli and DEVICE_PROXY_URL

### Type

Documentation

### Context

Doing this because I think just having easy access to a list of all existing super useful env variables, (in my case the `SKIP_ONBOARDING`) could **save hours** at least to people who are new to this codebase and don't know all the magic tricks to make the debugging experience confortable.

**Right now this documentation exists on Confluence so it's pretty much impossible to discover.**

Don't hesitate to commit on this PR if you find a better wording or more complete explanations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ easy data setting for development.
 
 Optional environment variables you can put in `.env`, `.env.production` or `.env.staging` for debug, release, or staging release builds respectively.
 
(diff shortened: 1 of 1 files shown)


gre (approved):
that's great. thanks for this.

PR #2067, opened 2021-12-20
fixed transaction broadcasting but still returning an error

IAmMorrow:
We have been facing a phantom bug for the last couple of months where transactions would be broadcasted but an error was being thrown to the dapp. This bug is now fixed.

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-8705

WebPlatformPlayer


Arnaud97234 (approved):

LFBarreto (approved):
:seal: :heavy_check_mark:

PR #2108, opened 2022-01-11
test pnpm

elbywan:
# DO NOT MERGE ⚠️ 

This is an experiment to ensure that pnpm can be used with LLM.


package.json
```diff
@@ -107,11 +110,15 @@
     "expo": "^43.0.1",
     "expo-barcode-scanner": "~11.1.2",
     "expo-camera": "~12.0.3",
+    "expo-image-loader": "^3.1.0",
+    "expo-modules-autolinking": "^0.5.5",
+    "expo-modules-core": "^0.6.4",
```
LFBarreto, line 115: :wave: 
why all the new deps?

elbywan: if I recall correctly these are transitive deps that are needed during the "pod install" command, but with pnpm without being dependencies they are placed in a nested directory which messes up the process

PR #2112, opened 2022-01-12
LL-8955 external contributions CI

valpinkman:
Adds a check to CI to run differently if coming from external fork or head repo

### Type

CI

### Context

[LL-8955]

[LL-8955]: https://ledgerhq.atlassian.net/browse/LL-8955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/bundle-app.yml
+++ b/.github/workflows/bundle-app.yml
@@ -17,18 +17,19 @@ jobs:
         with:
           access_token: ${{ secrets.GITHUB_TOKEN }}
 
-  check-org-member:
+  check-if-fork:
+    needs: [cancel-previous]
     runs-on: ubuntu-20.04
+    outputs:
+      forked: ${{ steps.forked.outputs.forked }}
     steps:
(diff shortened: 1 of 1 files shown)


elbywan (approved):

.github/workflows/bundle-app.yml
```diff
@@ -67,6 +68,7 @@ jobs:
         if: always()
         with:
           node-version: 14.x
+          cache: yarn
```
elbywan, line 71: Maybe yarn should be installed before this action runs?

elbywan (changes requested):
Wait I think I approved too quickly, I think there is an issue in the order of the steps in the action file 🤭

.github/workflows/bundle-app.yml
valpinkman: yup

elbywan (approved):
👌

PR #2117, opened 2022-01-13
LL-8962 Drop jcenter to fix android builds

juan-cortes:
When trying to build yesterday I noticed all  dependencies that were tried to be fetched from jcenter were failing, this proposes a workaround by following the alternatives each of the failing dependencies were proposing. In particular `com.redmadrobot:input-mask-android` which is used internally by our direct dependency `react-native-text-input-mask` needs to be resolved via a configuration change in the app.build, and the `react-native-fingerprint-scanner` we are pointing at a specific commit now since the library is no longer maintained and doesn't release.

We should drop these two dependencies if they don't solve the jcenter issue soon, but this is a solution in the meantime that shouldn't introduce any security vulnerabilities, the specific commit can't change, the rewrite points to their github, specific tagged version.

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-8962

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,7 +13,6 @@ buildscript {
     repositories {
         google()
         mavenCentral()
-        jcenter()
     }
     dependencies {
         classpath("com.android.tools.build:gradle:4.2.1")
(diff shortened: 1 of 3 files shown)


valpinkman (approved):
Ok as a temporary workaround

PR #2119, opened 2022-01-14
LL-8966 Plugin autoupdate feat

juan-cortes:
Force the usage of latest plugin/app versions when interacting with live apps, also set up a firmware version wall, allowing only devices with the latest firmware version to sign transactions within the discover section. I still think this is a bad idea, for the record.
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-8966

There's a test plan on Jira for this, just as it has for LLD.


imqdee (approved):
LGTM 🤷

PR #2120, opened 2022-01-14
LL-8570 - Settings page final part

nparigi-ledger:
Rebranding V3 - Setting page - Accounts, About, Help, Exp features, Debug, Ressources

https://user-images.githubusercontent.com/89014981/149811483-9007fa28-644d-4a75-a25c-02daadb39c66.mp4

:warning:
~~Merge https://github.com/LedgerHQ/ui/pull/162 https://github.com/LedgerHQ/ui/pull/155 https://github.com/LedgerHQ/ui/pull/154 before~~
Need these commits to be release to UI before this PR can be builded and marged

Note: "Set up a device" is handled by the Onboarding part, i doesn't work since  i had to use SKIP_ONBOARDING=1 because of misc onboarding issues on V3 (it's just a navigate so it should work nonetheless)

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -689,7 +689,7 @@
     }
   },
   "reset": {
-    "title": "Uninstall then reinstall ledger live",
+    "title": "Uninstall then reinstall Ledger Live",
     "description": "Please uninstall then reinstall the app on your phone to delete Ledger Live data, including accounts and settings.",
     "button": "Reset"
   },
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -689,6 +689,7 @@
     }
   },
   "reset": {
+    "title": "Uninstall then reinstall ledger live",
     "description": "Please uninstall then reinstall the app on your phone to delete Ledger Live data, including accounts and settings.",
     "button": "Reset"
   },
--- a/src/screens/Settings/About/AppVersionRow.tsx
+++ b/src/screens/Settings/About/AppVersionRow.tsx
@@ -1,0 +1,24 @@
+import React from "react";
+import VersionNumber from "react-native-version-number";
+import { Trans } from "react-i18next";
+import { Text } from "@ledgerhq/native-ui";
+import cleanBuildVersion from "../../../logic/cleanBuildVersion";
+import SettingsRow from "../../../components/SettingsRow";
+
+function AppVersionRow() {
+  const { appVersion, buildVersion } = VersionNumber;
+  const version = `${appVersion || ""} (${cleanBuildVersion(buildVersion) ||
+    ""})`;
+  return (
+    <SettingsRow
+      event="AppVersionRow"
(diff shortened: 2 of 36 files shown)


src/locales/en/common.json
```diff
@@ -689,6 +689,7 @@
     }
   },
   "reset": {
+    "title": "Uninstall then reinstall ledger live",
```
juan-cortes, line 692:
```suggestion
    "title": "Uninstall then reinstall Ledger Live",
```
I would assume this is what we expect, we never lowercase ledger

src/screens/Settings/About/AppVersionRow.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import VersionNumber from "react-native-version-number";
+import { Trans } from "react-i18next";
+import { Text } from "@ledgerhq/native-ui";
+import cleanBuildVersion from "../../../logic/cleanBuildVersion";
+import SettingsRow from "../../../components/SettingsRow";
+
+function AppVersionRow() {
+  const { appVersion, buildVersion } = VersionNumber;
+  const version = `${appVersion || ""} (${cleanBuildVersion(buildVersion) ||
```
juan-cortes, line 10: Since we are already using the same thing on the `useExportLogs` hook, consider extracting the logic of the clean app version into something reusable for both cases

src/screens/Settings/Resources.tsx
```diff
@@ -0,0 +1,83 @@
...
+  NewsMedium,
+  FacebookMedium,
+  TwitterMedium,
+  GithubMedium,
+  ActivityMedium,
+} from "@ledgerhq/native-ui/assets/icons";
+import SettingsNavigationScrollView from "./SettingsNavigationScrollView";
+import SettingsCard from "../../components/SettingsCard";
+
+const Resources = () => {
```
juan-cortes, line 16: Why aren't these urls in the urls file? I understand you just copied this but, perhaps, it's a good time to put the urls there too 🤔

juan-cortes (changes requested):

src/screens/Settings/Resources.tsx
nparigi-ledger: Indeed, and not too difficult to do suppose, i will check this

src/locales/en/common.json
nparigi-ledger: The whole text is used in uppercase but yes we never know, thanks.

src/screens/Settings/About/AppVersionRow.tsx
nparigi-ledger: Good idea, done

src/screens/Settings/Resources.tsx
nparigi-ledger: Done

nparigi-ledger:
@juan-cortes Thanks for your feedbacks, it's fixed

LFBarreto (approved):

PR #2123, opened 2022-01-17
LL-8449 - Fix segment using last seed device selector

tlaforge-ledger:
### Type

Bug Fix

### Context
https://ledgerhq.atlassian.net/browse/LL-8449

Events sending device information miss the informations and so on, the app to track events get some undefined values caused by those empty values.

The case to test are :
- try seeing what is sent on the events deviceVersion and modelId if I never connect a device
- try seeing what is sent on the events deviceVersion and modelId if I pair a BLE device and skip genuine check
- try only accessing the manager with a plugged in device and check deviceVersion and modelId

It should be missing on develop branch and fixed here if you have a last seen device defined.

we are fixing by doing :
- BLE users that for whatever reason failed to pass the genuine check. We offer a “skip genuine” check, which will write to the known device list but without the device info. This user will continue to report undefined forever since we never update that value, even if accessing the manager.
- Users who have only users a device through USB OTG, we are currently not updating the device information from these connections either. Again, even if they access the manager, these users will continue to report undefined, forever, unless they connect a BLE device at some point and pass the genuine check.

image

### DIFF
--- a/src/analytics/segment.js
+++ b/src/analytics/segment.js
@@ -17,6 +17,7 @@ import getOrCreateUser from "../user";
 import {
   analyticsEnabledSelector,
   languageSelector,
+  lastSeenDeviceSelector
 } from "../reducers/settings";
 import { knownDevicesSelector } from "../reducers/ble";
 import type { State } from "../reducers";
@@ -33,7 +34,8 @@ const extraProperties = store => {
   const { localeIdentifier, preferredLanguages } = Locale.constants();
   const language = languageSelector(state);
   const devices = knownDevicesSelector(state);
-  const lastDevice = devices[devices.length - 1];
+  
+  const lastDevice = lastSeenDeviceSelector(state) || devices[0];
   const deviceInfo = lastDevice
     ? {
         deviceVersion: lastDevice.deviceInfo?.version,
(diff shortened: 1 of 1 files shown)


src/analytics/segment.js
```diff
@@ -33,15 +34,16 @@ const extraProperties = store => {
   const { localeIdentifier, preferredLanguages } = Locale.constants();
   const language = languageSelector(state);
   const devices = knownDevicesSelector(state);
-  const lastDevice = devices[devices.length - 1];
+  
+  const lastDevice = lastSeenDeviceSelector(state) || devices[0];
```
juan-cortes, line 38: It should be `devices[devices.length - 1]` not `devices[0]` no? Otherwise we are getting the first one as a fallback

src/analytics/segment.js
```diff
@@ -33,15 +34,16 @@ const extraProperties = store => {
...
   const deviceInfo = lastDevice
     ? {
         deviceVersion: lastDevice.deviceInfo?.version,
         appLength: lastDevice.appsInstalled,
         modelId: lastDevice.modelId,
       }
     : {};
-
+
```
juan-cortes, line 46: Run prettier, I'm pretty sure these white spaces will be removed

juan-cortes (changes requested):

src/analytics/segment.js
tlaforge-ledger: I didn't commit it. 100%

juan-cortes (approved):
lgtm

OlivierFreyssinet (approved):
LGTM 👍

imqdee (approved):

PR #2125, opened 2022-01-18
fix podfile.lock for release

valpinkman:


juan-cortes (approved):
No idea why some deps go down instead of up though

IAmMorrow (commented):
lgtm

IAmMorrow (approved):

PR #2137, opened 2022-01-21
more generic condition and check for dependanbot

valpinkman:
allows dependabot to run as external contributor


elbywan (approved):
LGTM 👌

PR #2143, opened 2022-01-24
LL-8926: Fix polkadot bigInt issue

henrily-ledger:
### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-8926

I was able to fix this issues by taking an old version of the util lib.
The part of the problem was about BigInt into a Math.pow by using the `**` operator in polkadot lib.
It's seems to be find for this version but I don't know if it will work for later version (spoiler 7.5.1 just released yesterday, and LLD doesn't work anymore with this version 😠 )

https://github.com/polkadot-js/common/issues/1262#issuecomment-983995745

Polkadot, probably would be nice to retest the flow, but add accounts was mainly impacted

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,9 @@
     "**/**/pbkdf2": "3.1.1",
     "**/**/**/pbkdf2": "3.1.1",
     "react-native-svg": "12.1.1",
-    "d3-array": "2.3.3"
+    "d3-array": "2.3.3",
+    "@polkadot/util": "8.0.6-8",
+    "@polkadot/util-crypto": "8.0.6-8"
   },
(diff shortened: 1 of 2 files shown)


gre (approved):

haammar-ledger (changes requested):
You also need to push `yarn.lock` righ?

gre:
oh yeah good point from @haammar-ledger

henrily-ledger:
@haammar-ledger you told me yesterday but I'm still sleeping, I'm on my way

haammar-ledger (approved):

PR #2146, opened 2022-01-24
LL-7745 keyboard now opens automatically on password confirmation

cgrellard-ledger:
[LL-4045]

### Type

Bug Fix : Previously, the keyboard didn't opened correctly on some android phones when confirming password, it is now fixed

There actually wasn't a lot of changes in the code, I copied the js file in a tsx file (which is the reason so many changes appeared) but only the lines 41 to 43 changed (I added a check to not toggle the keyboard when already opened)

[LL-4045]: https://ledgerhq.atlassian.net/browse/LL-4045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/PasswordInput.tsx
+++ b/src/components/PasswordInput.tsx
@@ -1,0 +1,152 @@
+import React, { useEffect, useState, useCallback, useRef } from "react";
+import { View, StyleSheet, TextInput } from "react-native";
+import Icon from "react-native-vector-icons/dist/Feather";
+import Touchable from "./Touchable";
+import { getFontStyle } from "./LText";
+import { withTheme } from "../colors";
+
+type Props = {
+  secureTextEntry: boolean;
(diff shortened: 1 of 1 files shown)


juan-cortes (changes requested):
I'm very confused, these _changes_ are essentially what I proposed (and merged) on https://github.com/LedgerHQ/ledger-live-mobile/pull/1922 three months ago, changes that got dropped by @LFBarreto on https://ledgerhq.atlassian.net/browse/LL-8308 and they are coming back here again?

Do you mind linking to the right task on Jira, and filling in the sections of the PR (how to test, what changed, etc). As it stands it's not reviewable for me.


image

PR #2147, opened 2022-01-25
[LL-7025] - Language discoverability fixes

OlivierFreyssinet:
- Fix "discoverability for new users" -> the list of languages that can be applied by default (depending on the language of the OS) is **now restricted to English, Russian and French**, as opposed to "all supported languages". This was a blurry part of the spec, now clarified w/ @ahadwan-ledger. The reason is a dependency on available languages in external resources (articles, customer support etc.)
- Fix "discoverability for existing users" -> if the user selected the language, **the language was applied but not saved** for later runs of the app. This bug was due to a change in the logic for setting the app's language. I missed this part in my previous PR on this topic.

### Context

[LL-7025]

- Discoverability for new users (fully supported languages):
    1. Uninstall LL
    2. Switch OS language to french or russian (fully supported languages)
    3. Install LL, launch it
    4. LL should be in the system's language
- Discoverability for new users (not fully supported languages):
    1. Uninstall LL
    2. Switch OS language to a language not fully supported (german for instance)
    3. Install LL, launch it
    4. LL should still be in the english

- Discoverability for existing users
   1. Uninstall LL
   2. Switch OS language to english
   3. Install LL, launch it, go through onboarding
   4. Quit LL, switch OS language to french or russian (resp. a not fully supported language like German)
   5. Launch LL
   6. LL should prompt a dialogue asking the user to switch the language to the OS language (resp. if OS language is not a fully supported language, nothing should happen)

[LL-7025]: https://ledgerhq.atlassian.net/browse/LL-7025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/CheckLanguageAvailability.js
+++ b/src/components/CheckLanguageAvailability.js
@@ -52,7 +52,7 @@ export default function CheckLanguageAvailability() {
       <Track
         onMount
         event={`Discoverability - Prompt - ${defaultLanguage}`}
-        eventProperties={defaultLanguage}
+        eventProperties={{ language: defaultLanguage }}
       />
       <BottomModal
         id="CheckLanguageAvailabilityModal"
--- a/src/components/CheckLanguageAvailability.js
+++ b/src/components/CheckLanguageAvailability.js
@@ -1,12 +1,10 @@
 // @flow
 
 import React, { useState, useCallback } from "react";
-import { useSelector } from "react-redux";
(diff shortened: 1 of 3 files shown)


MortalKastor (approved):

juan-cortes (approved):
lgtm

PR #2166, opened 2022-02-02
LL-9182: reduce data store in dbs (proposal)

henri-ly:
### Type

BugFix + Improvement (probably)

### Context

https://ledgerhq.atlassian.net/browse/LL-9182

With the new BTC JS implementation we encounter a very large amount of data for a single account.
In Android we are not able to store more than 2 Mo per row dues to SQLite limitation.

> AsyncStorage for Android uses SQLite for storage backend. While it has its own size limits, Android system also have two known limits: total storage size and per-entry size limit.
Total storage size is capped at 6 MB by default. You can increase this size by specifying a new size using feature flag.
Per-entry is limited by a size of a WindowCursor, a buffer used to read data from SQLite. Currently it's size is around 2 MB. This means that the single item read at one time cannot be larger than 2 MB. There's no supported workaround from AsyncStorage. We suggest keeping your data lower than that, by chopping it down into many entries, instead of one massive entry. This is where multiGet and multiSet APIs can shine

https://react-native-async-storage.github.io/async-storage/docs/limits

So as they recommended I tried to compress JSON of large data and then split them into multiple part, not the sexiest way to fix it tho.
I'm up for any recommandation :)

EDIT: We remove compression because it cost a lot of time, we store bigger data raw.

Part that rely on device storage. A lot of part in the app :(

test to install this new version other another app having btc accounts with lots of tx
on a 2.36.x version probably
check if the app boots with no issue and the data is migrated correctly
after a sync of the accounts stop the app and relaunch it
the app should reboot correctly without issue

this should cover all cases of old phones having to migrate data, phones with already migrated accounts but not big enough to kill the db and the rest.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -355,7 +355,7 @@ PODS:
   - RNAnalytics (1.5.0):
     - Analytics
     - React-Core
-  - RNCAsyncStorage (1.12.1):
+  - RNCAsyncStorage (1.15.17):
     - React-Core
   - RNCClipboard (1.5.1):
     - React-Core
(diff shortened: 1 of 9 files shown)


hzheng-ledger (approved):
Looks good to me. Good fix!

gre (approved):

LFBarreto (approved):
:rocket:

LFBarreto:
@henrily-ledger could you sign all your commits pls :pray:

LFBarreto (commented):
Could you add a try catch around the Db read in order to unlock the boot of currently blocked apps :pray:

juan-cortes (approved):

PR #2172, opened 2022-02-03
LIVE-559 Ledger Nano S Plus support in Device Actions + other assets

juan-cortes:
> #### Note: The huge number of lines deleted is due to dropping the html files for the lottie animations. We no longer need them since we have a viewer now.
--------

This brings asset support for the new product that was announced. It should cover everything from the onboarding, device actions, and manager assets. Due to the limited number of devices of this model I have added a few env variables to allow for easier testing. I've added indications on the Test plan below.

Group

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LIVE-559

I've added a few env variables but this means that you can't just:
`SHOW_NANOSP` If we don't pass this flag, the device selection screen for the onboarding flow will only show the nano s and the nano x. We will remove this check when we officially release a version supporting this device.
`OVERRIDE_MODEL_ID` In case you want to test a build simulating a specific device, you can pass the model id as an env. With the values `nanoS, nanoSP or nanoX`

Since you are probably going to use the artifacts and not build the file, I've added a new screen in the debug menu to be able to add some env at runtime. To access this feature navigate to Debug/Mock & Test/ Debug Env, and set the env you need above. Click on the button a few times and hopefully, it will work as expected 🤞🏼


src/screens/Onboarding/steps/deviceSelection.js
```diff
@@ -31,14 +47,14 @@ function OnboardingStepDeviceSelection({ navigation }: *) {
...
-      {Object.keys(deviceIds).map((deviceId, index) => (
+      {Object.keys(deviceIds).map((deviceId: any, index) => (
         <Touchable
           key={deviceId + index}
           event="Onboarding Device - Selection"
           eventProperties={{ deviceId }}
           testID={`Onboarding Device - Selection|${deviceId}`}
           style={[styles.deviceButton, { backgroundColor: colors.lightLive }]}
-          onPress={() => next(deviceId)}
+          onPress={() => next("nanoSP")}
```
tlaforge-ledger: I think you have some testing stuff

juan-cortes: Indeed, good catch

nparigi-ledger (approved):

grsoares21 (approved):

tlaforge-ledger (approved):

PR #2190, opened 2022-02-11
LL-8639 Handle OS theme properly

juan-cortes:
Offer the user an option to follow the system theme instead of overriding their choice when the OS theme changes. Default to that instead of dark. Keep in mind that there are three values for the operating system theme. If you've never set it before it will be `null`, if I've set it to dark before it will be `dark` but if I disable dark it will be `light`.

https://user-images.githubusercontent.com/4631227/153578351-b885e192-b121-44a8-8ae8-977a9c93f5b6.mov

### Type

Bug Fix

### Context
https://ledgerhq.atlassian.net/browse/LL-8639

- When the app theme is "system" it should follow the OS selection
- When the app theme is "light" it should ignore the OS selection and be light.
- When the app theme is "dark" it should ignore the OS selection and be dark.

### DIFF
--- a/src/actions/settings.js
+++ b/src/actions/settings.js
@@ -25,7 +25,7 @@ type SetExchangePairs = (
   }>,
 ) => *;
 
-export type Theme = "light" | "dark";
+export type Theme = "system" | "light" | "dark";
 
 export const setExchangePairsAction: SetExchangePairs = pairs => ({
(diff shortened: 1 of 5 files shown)


LFBarreto (approved):

LFBarreto (dismissed):
@juan-cortes could you cherry pick your changes and target the release branch pls :pray:

PR #2197, opened 2022-02-14
LIVE-926 Disable closing modals while inline install is happening

juan-cortes:
Since the introduction of the feature [back in April 2021](https://github.com/LedgerHQ/ledger-live-mobile/pull/1679#pullrequestreview-646260570) @aulric-ledger noticed that the inline installation could be stopped midway by either closing the modal or tapping outside of it. At the time we couldn't think of an easy solution to prevent this but I see no reason why this wouldn't work. 

https://user-images.githubusercontent.com/4631227/153897273-5afa0dc7-c6d1-4bef-a649-0df265b24358.mov

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-926

- Enter a device action flow without the appropriate application
- Notice that the bottom modal can be closed at any given time
- During inline installation the X that closes the modal should not be present
- Clicking on the backdrop should not close the modal
- Clicking _back_ on the device (Android) should not close the modal

Completing the installation or failing (device disconnection for instance) should also unlock the modal and allow closing it. In other words, it should only remain locked during the inline installation and any other step that was locked before.

### DIFF
--- a/src/actions/appstate.js
+++ b/src/actions/appstate.js
@@ -18,3 +18,6 @@ export const syncIsConnected = (isConnected: boolean | null) => (
 export const setHasConnectedDevice = (hasConnectedDevice: boolean) => (
   dispatch: *,
 ) => dispatch({ type: "HAS_CONNECTED_DEVICE", hasConnectedDevice });
+
+export const setModalLock = (modalLock: boolean) => (dispatch: *) =>
+  dispatch({ type: "SET_MODAL_LOCK", modalLock });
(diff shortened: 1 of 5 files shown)


tlaforge-ledger (approved):

thomasrogerlux (approved):

OlivierFreyssinet (approved):
looks good to me 👌

LFBarreto (approved):

PR #2203, opened 2022-02-15
LIVE-1193 Fix app sizes in Manager

juan-cortes:
image

The latest version of the Ledger Nano S Plus allows for block sizes that go below 1kB. For apps that are below that size product wants us to display the size in kB with one decimal point. On that line, we changed the units to be kB and MB because it seems those are the recommended units.
### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1193

- Having a Ledger Nano S Plus on the latest greatest version
- Use provider 12
- Enter the manager and check that the sizes below one kB are shown as 0.X kB and not in bytes.

### DIFF
--- a/src/components/ByteSize.js
+++ b/src/components/ByteSize.js
@@ -10,7 +10,7 @@ const sizes = ["bytes", "kbUnit", "mbUnit"];
 const ByteSize = ({
   value,
   deviceModel,
-  decimals = 0,
+  decimals = 2,
   firmwareVersion,
   formatFunction,
 }: {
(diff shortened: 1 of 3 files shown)


tlaforge-ledger (approved):

LFBarreto (approved):

OlivierFreyssinet (approved):

thomasrogerlux (approved):
Is it normal that the yarn lockfile changed? Can't test without S Plus but code looks fine

PR #2214, opened 2022-02-16
LL-1330 Onboarding animations for LNS+

juan-cortes:
Since the change to dark mode as a default for the applications we noticed a bug in the onboarding animations that made the cable for the LNS+ not visible in dark mode. In order to fix this we had to refactor the code a little bit in preparation to having new lottie animations that are different between dark and light mode (that's why the LNS and LNX animations are repeated currently) which will come soon, but not right now.

image
This animation used to not have a visible cable

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1330

Following the test plan from https://github.com/LedgerHQ/ledger-live-mobile/pull/2172 is probably the best option, considering the impact is the same (only the onboarding part) and the tool for testing is also the same.


tlaforge-ledger (approved):

grsoares21 (approved):
LGTM

PR #2228, opened 2022-02-21
LIVE-1392 Show NanoSP by default

juan-cortes:
Enable the Ledger Nano S Plus by default on the application, without requiring the `SHOW_NANOSP` flag like we needed up until now. This means the device should be available during onboarding and detected when plugged in (this last part was already the case).

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1392

- Make sure the app shows the device without needing the flag. All other functionality was tested in the specific PRs.


juan-cortes:
QA Ok pending dev review

gre (approved):

thomasrogerlux (approved):

PR #2250, opened 2022-03-01
LIVE-1533 - Update Firebase environments (prod/staging/dev)

thomasrogerlux:
[LIVE-1533]

### Type

On Desktop and Android, we have different configuration of Firebase depending on the context (prod/staging/dev), this PR setup the same system on iOS.

Edit: also this is the final credentials update (had to do a few changes Firebase side because of the way we handle bundle ids iOS side).

### Context

There's three Firebase projects, one for production, one for staging and one for dev. We need to load different configuration on app startup depending on the environment.

iOS setup

[LIVE-1533]: https://ledgerhq.atlassian.net/browse/LIVE-1533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OlivierFreyssinet (approved):
👌

ychen-ledger (approved):
Validated with: 
- Pixel 4 Android 12 / iPhone 11 iOS 14.7
- LLM 2.39.1 - debug app
-  firebase - activated - deactivated

PR #2254, opened 2022-03-01
v2.39.1 on develop

lambertkevin:


juan-cortes:
Lacks this https://github.com/LedgerHQ/ledger-live-mobile/pull/2255 and the related live common bump if we want to cover usb org nano s + unseeded

PR #2258, opened 2022-03-03
Text-encoding-polyfill

henri-ly:
### Type

Bugfix for the implementation of cosmos js

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1577

Should take care of all the app... Because it's a global polyfill

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,9 @@ import "react-native-gesture-handler";
 // https://github.com/facebook/react-native/issues/23922
 import "react-native-url-polyfill/auto";
 
+// cosmjs use TextEncoder that's not available in React Native but on Node
+import "text-encoding-polyfill";
+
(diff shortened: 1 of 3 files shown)


gre:
i studied a bit the impact by a `ack TextEncoder` in node_modules and I can see all these coin integration are possibly affected because indirectly use a lib that use TextEncoder:
- polkadot
- ripple-lib
- algorand
- solana
- tezos (js rework)

Arnaud97234 (dismissed):
Add account fails for solana

ghost (approved):

mcoursier-ledger (approved):
I verified the coins listed on my end and did not find any issue on iOS and Android.
But I guess we should pay a special attention on this change during our regression testing phase for the next release.

gre (approved):

PR #2260, opened 2022-03-03
LIVE-1599 - Fix bottom drawer inconsistencies

OlivierFreyssinet:
Removing extra ❌ close buttons in bottom drawers as it is already present in the new `BottomDrawer` component from the UI lib.

- I went through all the components using `<BottomModal />` and looked for the ones using the close icon.
- This doesn't include removal of textual "close" CTAs as these might have to be handled differently case by case.
- Regarding the removal of the close icon in `DeviceActionModal`, that button was wrapped in a component that would hide the button in case the modal is supposed to be locked (`modalLock`). I pulled up the logic for locking a bottom in our `BottomModal` component used everywhere. This logic depends on a [new prop for the modals](https://github.com/LedgerHQ/ui/pull/174) in the UI lib. Implementation details here: https://github.com/LedgerHQ/ledger-live-mobile/pull/2260/commits/36d88b921649e6a6c5b4dadf55fe6db73df95e7b and here https://github.com/LedgerHQ/ledger-live-mobile/pull/2260/commits/9887418040c545dcbacf908ecef127483a3d97d9

⚠️ dependency on this UI lib PR https://github.com/LedgerHQ/ui/pull/174 (last point mentioned above)

#### Example:
Combined

### Type

Bug fix

### Context

[LIVE-1599]

[LIVE-1599]: https://ledgerhq.atlassian.net/browse/LIVE-1599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


LFBarreto (approved):

ofreyssinet-ledger:
I'm stupid I forgot to upgrade the UI lib

PR #2261, opened 2022-03-03
LIVE-1601 Use FastImage component for all platforms

grsoares21:
Currently we’re using different components for displaying NFT Images on iOS and Android. The component being used on Android has worse performance, this is due to an inability to handle certain types of images on that platform. However, with the new image processing service, all images are being converted to a supported format, this fixes this problem and we can use the fast component for both Android and iOS.

### Type
UX Improvement (performance)

### Context
[LIVE-1601]

All NFT pages (Gallery and NFT viewer) should be working seamlessly.

[LIVE-1601]: https://ledgerhq.atlassian.net/browse/LIVE-1601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Nft/NftImage.js
+++ b/src/components/Nft/NftImage.js
@@ -20,24 +20,15 @@ const ImageComponent = ({
   resizeMode: string,
   onLoadEnd: () => *,
   onLoad: () => *,
-}) =>
-  Platform.OS === "android" ? (
-    <Image
-      style={style}
-      resizeMode={resizeMode}
-      source={source}
-      onLoad={onLoad}
-      onLoadEnd={onLoadEnd}
-    />
-  ) : (
-    <FastImage
-      style={style}
-      resizeMode={FastImage.resizeMode[resizeMode]}
-      source={source}
-      onLoad={onLoad}
-      onLoadEnd={onLoadEnd}
-    />
-  );
+}) => (
+  <FastImage
+    style={style}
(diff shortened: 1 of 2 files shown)


nparigi-ledger (approved):

OlivierFreyssinet (commented):
What happens when an image type isn't supported like it was the case on Android? Does it just stay blank or is there an error?

OlivierFreyssinet (approved):

juan-cortes (approved):

lambertkevin:
> What happens when an image type isn't supported like it was the case on Android? Does it just stay blank or is there an error?

The problem was actually not compatibility before. Using FastImage was actually the best way to support GIFs, but we decided not to support it because we still had major performance issues and actually supporting GIFs on Android was making it worse. 
Using the basic Image component made the GIF a static image and was making the UI less laggy in the gallery view or the collection view. 
Now that we have some image processing directly in our backend, we're served with a static image for "preview" which means that we can afford to support GIFs everywhere since the pages with a lot of NFTs are going to use the static preview, and only the NFT Viewer (so only 1 NFT) is going to use an (potential) animated image.

ofreyssinet-ledger:
Nice, thanks for clarifying!

PR #2267, opened 2022-03-04
LIVE-1605 Fix receive flow loop

juan-cortes:
Bad hook dependencies on the `useEffect` that was triggering the address confirmation were causing a re-trigger whenever the affected account was updated in the background. Basically if you were in the confirmation screen long enough, your account would sync and in doing so it would re-trigger the hook logic. Removing the dependencies and adding the extra condition of not having completed the confirmation fixes this, since a retry basically navigates back.

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1605

- Go through a receive flow
- Confirm the address on your device
- Wait, if nothing happens, we're good :trollface:

### DIFF
--- a/src/screens/ReceiveFunds/03-Confirmation.js
+++ b/src/screens/ReceiveFunds/03-Confirmation.js
@@ -160,13 +160,13 @@ export default function ReceiveConfirmation({ navigation, route }: Props) {
   useEffect(() => {
     const device = route.params.device;
 
-    if (device) {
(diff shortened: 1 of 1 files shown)


juan-cortes:
QA OK ✅

ghost (approved):

thomasrogerlux (approved):

elbywan (approved):
👌

PR #2276, opened 2022-03-08
fix: allow install for hash commit version

valpinkman:
Allows for hash commit version in package.json to install correctly

### Type

Improvement

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,7 +19,6 @@ jobs:
       - uses: actions/setup-node@master
         with:
           node-version: 14.x
-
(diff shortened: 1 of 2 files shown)


OlivierFreyssinet (commented):
What exactly is breaking without `--network-concurrency 1` ?

valpinkman:
> 



> What exactly is breaking without `--network-concurrency 1` ?

@ofreyssinet-ledger basically, whenever you install a hash commit of LLC, the prepare script in package.json is run, which makes it runs 2 yarn side to side and thus breaks the install process

OlivierFreyssinet (approved):

PR #2277, opened 2022-03-08
LIVE-1443 LIVE-1170 react-native-fact-crypto added

hzheng-ledger:
depends on live-common develop

### Jira tickets
- https://ledgerhq.atlassian.net/browse/LIVE-1443
- https://ledgerhq.atlassian.net/browse/LIVE-1170

### Acceptance criterias (for QA)

bot is overall passing which give us confidence enough to have a ready to QA build. even if the impact is in live-common, it's not necessary to test on LLD to me with the overall bot assessment + if we focus on confirming mobile is good. but happy to discuss more.

**Please also beware of this impact:**
- as we removed a library for zcash and changed a bit our logic of ZEC, there is something specific done on this PR about ZCash so I would say QA should also extra check carefully on ZCash too.
- most of the changes are impacting synchronisation (and therefore add accounts). but it's good to do at least one transaction to confirm that we didn't break that, maybe interesting also to see that the "receiving account" receives a new tx, as it's part of sync.

#### Non-reg
- Can build on iOS and Android
- Add and sync all BTC like accounts
- Incremental sync: do a transaction, make sure account is correctly updated
- Non-reg: import a few accounts from live desktop


ghost (approved):

gre (approved):

gre:
will need some extra logic to connect it to the "set implementation" right?

src/logic/secp256k1.js
```diff
@@ -0,0 +1,18 @@
...
+      Buffer.from(tweak).toString("base64"),
+    ),
+    "base64",
+  );
+  // buffer is a uncompressed public key, we should return a compressed public key. refer to https://bitcoin.stackexchange.com/questions/3059/what-is-a-compressed-bitcoin-key
+  return Promise.resolve(
+    Buffer.concat([
+      Buffer.from(buffer[64] % 2 === 0 ? "02" : "03", "hex"),
+      buffer.slice(1, 33),
+    ]).toJSON().data,
```
gre, line 16: what is the type of this? `toJSON().data` feels weird, but i'm not sure what is the returned type here. is it a plain JS array? we can't return the buffer as is? thanks

ghost (changes requested):
CI error (in CI tests):
```
 /home/runner/work/ledger-live-mobile/ledger-live-mobile/src/live-common-setup.js
Error:   15:38  error  Unable to resolve path to module '@ledgerhq/live-common/lib/families/bitcoin/wallet-btc/crypto/secp256k1'  import/no-unresolved

✖ 1 problem (1 error, 0 warnings)
```

src/logic/secp256k1.js
hzheng-ledger: it is a plain JS array.
I've update the return type to Promise<number[]> now on LLC side

hzheng-ledger:
> CI error (in CI tests):
> 
> ```
>  /home/runner/work/ledger-live-mobile/ledger-live-mobile/src/live-common-setup.js
> Error:   15:38  error  Unable to resolve path to module '@ledgerhq/live-common/lib/families/bitcoin/wallet-btc/crypto/secp256k1'  import/no-unresolved
> 
> ✖ 1 problem (1 error, 0 warnings)
> ```

this is because that the change of LLM here depends on the PR from LLC. I just updated the package.json so that it points to the corresponding commit hash. It should fix the problem. Thanks

henrily-ledger:
As an update, it doesn't seems to work on iOS.

The problem probably come from NativeModules that can't be found in 'react-native'. So you will get an `undefined` in the index.js of the package.

```
import NativeModules from 'react-native'

const { RNSecp256k1, RNSecp256k1Ext} = NativeModules
```

I have no idea why, try to dig for a solution but wasn't be able to find any yet.

gre:
i don't really understand why the line in our own podfile is needed:

```
  pod 'RNSecp256k1', :path => '../node_modules/@ledgerhq/react-native-secp256k1/ios'
  ```
  
  as it's supposely covered by 
 
 ```
  config = use_native_modules!
```

if i'm correct. like for most RN libs, we didn't had a dedicated pod line

gre:
I don't see the RNSecp256k1* pods appearing in the "Products" folder of "Pods" in XCode, so it's a problem

gre:
in the Pods/RNSecp256k1/ folder, we can notice the .h and .m don't even appear like other RN files.

i have changed that lib podspec to 
```
  s.source_files  = "*.{h,m}"
```

and after a yarn pod, it helps finding them...

i feel that library isn't properly configured 🤔

gre (approved):

ghost (approved):

PR #2294, opened 2022-03-15
Remove headerMode:"float" EVERYWHERE as it's breaking navigation screens

OlivierFreyssinet:
#### Before
Combined

#### After

adb-screencap(6)

### Type

Bug fix


LFBarreto (approved):
:sparkles: joy

PR #2303, opened 2022-03-17
Fixes Tezos JS import

gre:
### Type

bugfix on tezos experimental

### Context

LIVE-1689

the impact is exclusively on Tezos and on mobile. It fixes the ability to do transactions after importing Tezos accounts from Ledger Live Desktop. you must remove previous accounts if you had them imported. It is to me not necessary to test on both iOS and Android, the bug was cross platform.

The fix was reproduced by the bot and then proven to be fixed by the bot in this PR: https://github.com/LedgerHQ/ledger-live-common/pull/1818

Builds

- https://github.com/LedgerHQ/ledger-live-build/actions/runs/2000458062
- https://github.com/LedgerHQ/ledger-live-build/actions/runs/2000457466

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2537,7 +2537,7 @@
 
 "@ledgerhq/live-common@https://github.com/LedgerHQ/ledger-live-common.git#develop":
   version "21.34.1-cosmos.4"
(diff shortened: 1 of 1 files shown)


gre:
fix is confirmed

PR #2310, opened 2022-03-21
Update latest live-common on develop

gre:
this is just like a /update-llc, it will update latest develop of live-common on develop of LLM.


thomasrogerlux (approved):

gre:
ok it's the same as https://github.com/LedgerHQ/ledger-live-mobile/pull/2304 , once tezos pr is validated i think we merge

PR #2328, opened 2022-03-24
LIVE-1748 Prevent crash when using proxy device

juan-cortes:
### Type

Bugfix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1748

- Start the app using a proxy device
- Try accessing the manager with that device
- See if it crashes or not, it shouldn't!

### DIFF
--- a/src/components/DeviceAction/getDeviceAnimation.js
+++ b/src/components/DeviceAction/getDeviceAnimation.js
@@ -84,8 +84,8 @@ const animations = {
     },
     bluetooth: {
       plugAndPinCode: {
-        light: null,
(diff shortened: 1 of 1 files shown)


gre (approved):

tlaforge-ledger (approved):

juan-cortes:
Nobody seems to be able to test this since we can't get the proxy to work at all on LLM.
Since this is essentially used for dev and @gre is off now, we can wait until he's back. I believe the fix is clear and should work so it should be harmless to merge.

LFBarreto (approved):

PR #2352, opened 2022-03-30
Stake button added for Ethereum Accounts

cgrellard-ledger:
-->


src/components/FabAccountButtonBar.tsx
```diff
@@ -84,7 +85,13 @@ function FabAccountButtonBar({
       const { navigationParams, confirmModalProps } = data;
       if (!confirmModalProps) {
         setInfoModalProps();
-        if (navigationParams) onNavigate(...navigationParams);
+        if (navigationParams) {
+          if (typeof navigationParams === "string") {
+            Linking.openURL(navigationParams);
```
LFBarreto, line 90: better use another field like deelink instead of having mixed type fields :pray:

LFBarreto (changes requested):

PR #2356, opened 2022-03-30
V3 - portfolio wallet value add time range

nparigi-ledger:
Screenshot_20220330-144629_LL  DEV

### Type

Bug fix

### Context

https://docs.google.com/spreadsheets/d/16kFSnt7z6p3ro1yAl09FIxsFMm_bkVm00mHgfFfD2fs/edit#gid=0

Portfolio page

### DIFF
--- a/src/components/Delta.tsx
+++ b/src/components/Delta.tsx
@@ -1,22 +1,29 @@
 import React, { memo } from "react";
 import { View, StyleSheet } from "react-native";
 import { Unit } from "@ledgerhq/live-common/lib/types";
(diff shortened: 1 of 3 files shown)


src/components/Delta.tsx
```diff
@@ -54,6 +61,7 @@ function Delta({ valueChange, percent, unit, style }: Props) {
           ) : percent ? (
             `${absDelta.toFixed(0)}%`
           ) : null}
+          {range && <> ({t(`time.${range}`)})</>}
```
LFBarreto, line 64: do you need the fragment here? <></> ?

PR #2220, opened 2022-02-18
Bugfix/LIVE-973 - Discreet mode restricted to portfolio & accounts screens

OlivierFreyssinet:
Discreet mode should only affect values displayed in the Portfolio and Account/Accounts screens.
I implemented a new Context `DiscreetModeContext` and its provider `DiscreetModeProvider` can be used to wrap entire screens that should be affected by the discreet mode. A HOC `withDiscreetMode` makes wrapping existing components even easier.

The specification isn't extremely detailed as it just says "only hides portfolio value and individual accounts value" but I've chosen these screens/components to be affected by the discreet mode:

**Account screens: (in `AccountsNavigator`)**
- `screens/Account`
- `screens/Accounts`
- `screens/Nft/NftGallery`
- `screens/Nft/Collection`

**Operation details screen:**
- `screens/OperationDetails`

**Portfolio screens:**
- `screens/Portfolio`
- `screens/Asset` the screen that shows up when pressing an asset in the portfolio "asset allocation" section.
- `screens/Distribution` the screen that shows up when pressing the "all" CTA in the portfolio "asset allocation" section.

In the following video I'm showing all of these screens in discreet mode, and at the end one screen (the buy flow / select account screen) to demonstrate that values outside of these screens aren't affected anymore by discreet mode.

https://user-images.githubusercontent.com/91890529/154706222-e40162d7-f45f-43e4-aed8-1673bbfb647f.mp4

### Type

Bug fix

### Context

[LIVE-973]

### Parts of the app affected / Test plan

[LIVE-973]: https://ledgerhq.atlassian.net/browse/LIVE-973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2365, opened 2022-03-30
Fix regression on discreet mode

OlivierFreyssinet:
This fixes a regression on the _selective discreet mode_ bug fix: discreet mode only applies to some screens/component, i.e. portfolio, distribution screen & account screen, instead of applying to all displayed currency values in the app.
This bug fix/feature had been [implemented here](https://github.com/LedgerHQ/ledger-live-mobile/pull/2220/) but only in js/jsx files while tsx files already existed for v3, so it was missing in v3.

https://user-images.githubusercontent.com/91890529/160873794-d558237c-9443-4570-9cb6-872a6ede026f.mp4

### Type

Bug fix / regression

### Context

v3 pre-release

Portfolio, distribution screen, account page. In general all amounts displayed in the app, but now discreet mode is _opt-in_ (from a dev perspective) so _by default_ (without this opt-in) all currency amounts displayed in the app will just be in clear text.

### DIFF
--- a/src/components/CounterValue.js
+++ b/src/components/CounterValue.js
@@ -35,6 +35,7 @@ type Props = {
   Wrapper?: React$ComponentType<*>,
   subMagnitude?: number,
   joinFragmentsSeparator?: string,
+  alwaysShowValue?: boolean,
 };
 
 export const NoCountervaluePlaceholder = () => {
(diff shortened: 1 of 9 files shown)


cgrellard-ledger (approved):
good

ThomasLaforge (approved):

LFBarreto (approved):
:see_no_evil:

PR #2380, opened 2022-04-01
Android - Allow cleartext communication by default in debug builds

OlivierFreyssinet:
**TL;DR: build configuration to allow all cleartext HTTP traffic in Android _debug_ builds, allowing devs to easily debug the app running on an emulator with a real nano (using the proxy), and allowing an easy wireless debugging experience.** 

On Android we need to whitelist IP adresses for which we want to allow cleartext HTTP (and not HTTPS) traffic. This happens in a `network_security_config.xml` resource file.

There are several issues with that for developer experience:
- The most important issue is with `ledger-live proxy`, if a dev wants to debug the Android app running in an emulator with a real device, the only way is to use that tool. Because of that whitelist, the dev will have to manually add the IP address given by the proxy CLI. 
- Another issue is to allow debugging wirelessly on a device. Even though there are other ways to do it efficiently (with ADB over wifi), React Native provides a way to do it by simply inputting in the app the IP of the Metro bundler server. Here again, it won't work unless the IP is added to the whitelist.

In both cases, the dev has to input an IP address manually in the configuration, **rebuild the android app** (first pain point), and also **make sure to not push these changes** to the security config (second pain point).

The solution here is to simply add a `network_security_config.xml` specific to debug builds (`debug` buildType) and that configuration will by default allow any cleartext HTTP traffic.

**⚠️ One small downside of this solution:** the debug environment will not reflect how the prod app behaves with HTTP traffic ... (it will not work by default in prod)

@grsoares21 you might be interested

### Type

DX improvement

### DIFF
--- a/android/app/src/debug/res/xml/network_security_config.xml
+++ b/android/app/src/debug/res/xml/network_security_config.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="true" />
+</network-security-config>


LFBarreto (approved):
🐸

PR #2414, opened 2022-04-11
Fixes Sentry support

gre:
This reboots the support of Sentry that was broken since Nov 2021.
Took the opportunity to update Sentry & switch to another project (ledger-live-mobile)
It also uses the latest API of sentry (routing instrumentation, Sentry.warp, ..)

BTW I didn't manage to understand why it was broken and I have only tested the integration locally on iOS and Android (not in production condition, in dev mode against a custom project on sentry, i was able to confirm errors reach the sentry backoffice) so we'll use future builds to confirm the integration is correctly working end2end.


OlivierFreyssinet (approved):
looks good to me 👍
glad we're taking care of this 😁

PR #2429, opened 2022-04-13
LIVE-1819 FTS Support

juan-cortes:
### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LIVE-1819

No image or video since we're just using the nano s illustrations and this is for development purposes.


juan-cortes:
Should be ready for testing again now that we solved the v3 dependency, @alalmi-ledger

alalmi-ledger (commented):
QA OK

src/screens/PairDevices/Scanning.tsx
```diff
@@ -32,11 +33,17 @@ export default function Scanning({ onTimeout, onError, onSelect }: Props) {
   const renderItem = useCallback(
     ({ item }) => {
       const knownDevice = knownDevices.find(d => d.id === item.id);
+      let modelId = "nanoX" as DeviceModelId;
+      if (item?.serviceUUIDs.length) {
```
lambertkevin, line 37: Shouldn't you add a `?` optionnal chaining on length ? If `serviceUUIDs` is undefined, should be problematic to get the length, isn't it ?

lambertkevin (approved):
Just 1 comment, but other than that LGTM 👌

PR #2442, opened 2022-04-19
Terms of use update popup

thomasrogerlux:
[LIVE-1914]

We wish to notify users properly when the Terms of Use gets updated, directly through Ledger Live. This PR add a new modal in the Portfolio component that will open when the `accepted` value from the terms context is false and the user isn't in the onboarding. Closing the modal accepts the new terms.

Screenshot from 2022-05-04 16-52-25

- Portfolio
- New `CheckTermOfUseUpdate` component
- Terms of uses

[LIVE-1914]: https://ledgerhq.atlassian.net/browse/LIVE-1914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ngoctuan8092 (approved):

ThomasLaforge (approved):
comparing dates is much better than the old comparison of string IMO. Good one !

cgrellard-ledger:
Do we want to display the close button in the top right corner ?

thomasrogerlux:
> Do we want to display the close button in the top right corner ? `Modal will be blocking user has to click on “Continue”` <= from the specs

It's removed, I need to update the screenshot in the comment

cgrellard-ledger (approved):

nparigi-ledger (approved):

nparigi-ledger:
Tested it by putting currentTermsRequired at a far away date.
The modal is popping without problem and the link is working.
Tested in spanish. The terms of use redirect to the spanish version of terms of use.
Tested in russian. As it doesn't seems to exist, it use the english version as fallback.

PR #2333, opened 2022-03-24
LIVE-1625 Add hide NFT collection feature

grsoares21:
This PR adds the possibility of hiding NFT collections from both the account page and the NFT Gallery page.
The hidden collections are visible in the account settings page and can be unhidden there.

https://user-images.githubusercontent.com/6013294/159961708-04afeea7-3f4a-45e4-8543-f53a42153125.mp4

### Type
Feature

### Context
[LIVE-1625]

### Parts of the app affected / Test plan
Account page, NFT Gallery and Setting page.
On an account page, long-press an NFT collection to show the hide menu.
On the NFT Gallery, click the 3 dots beside the collection title to show the hide menu.
Go to settings -> accounts -> hidden NFT collections to unhide the collections.
Hidden collections should not be visible in the account page nor the NFT gallery.

[LIVE-1625]: https://ledgerhq.atlassian.net/browse/LIVE-1625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2452, opened 2022-04-26
Feat/live 1625 hide nft collection v3 rebased

lambertkevin:
### Copy of #2333 

This PR adds the possibility of hiding NFT collections from both the account page and the NFT Gallery page.
The hidden collections are visible in the account settings page and can be unhidden there.

https://user-images.githubusercontent.com/6013294/159961708-04afeea7-3f4a-45e4-8543-f53a42153125.mp4

### Type
Feature

### Context
[LIVE-1625]

Account page, NFT Gallery and Setting page.
On an account page, long-press an NFT collection to show the hide menu.
On the NFT Gallery, click the 3 dots beside the collection title to show the hide menu.
Go to settings -> accounts -> hidden NFT collections to unhide the collections.
Hidden collections should not be visible in the account page nor the NFT gallery.

[LIVE-1625]: https://ledgerhq.atlassian.net/browse/LIVE-1625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Touchable.js
+++ b/src/components/Touchable.js
@@ -63,23 +63,16 @@ export default class Touchable extends Component<
   };
 
   onLongPress = async () => {
-    const { onLongPress, event, eventProperties } = this.props;
-    if (!onLongPress) return;
-    try {
+      const { onLongPress, event, eventProperties } = this.props;
+      if (!onLongPress) return;
       if (event) {
         track(event, eventProperties);
       }
+      
       const res = onLongPress();
-      if (res && res.then) {
-        // it's a promise, we will use pending/spinnerOn state
-        this.setState({ pending: true });
-        await res;
-      }
-    } finally {
-      if (!this.unmounted) {
-        this.setState(({ pending }) => (pending ? { pending: false } : null));
-      }
-    }
+      this.setState({ pending: true });
+      await res;
+      this.setState({ pending: false });
   };
 
   render() {
--- a/src/components/Touchable.js
+++ b/src/components/Touchable.js
@@ -18,6 +18,7 @@ type Props = {
   // will wait the promise to complete before enabling the button again
   // it also displays a spinner if it takes more than WAIT_TIME_BEFORE_SPINNER
   onPress: ?() => ?Promise<any> | void,
+  onLongPress: ?() => ?Promise<any> | void,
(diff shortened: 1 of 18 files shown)


src/components/Touchable.js
```diff
@@ -61,6 +62,26 @@ export default class Touchable extends Component<
     }
   };
 
+  onLongPress = async () => {
```
juan-cortes, line 65: Not sure about this @grsoares21 seems  overkill

src/components/Touchable.js
```diff
@@ -61,6 +62,26 @@ export default class Touchable extends Component<
...
+        // it's a promise, we will use pending/spinnerOn state
+        this.setState({ pending: true });
+        await res;
+      }
+    } finally {
+      if (!this.unmounted) {
+        this.setState(({ pending }) => (pending ? { pending: false } : null));
+      }
+    }
+  };
```
juan-cortes, line 83: @grsoares21 seems like this block is a bit overkill, can't we just do this and if it's not a promise then it should be fast enough to not be noticeable?? (Haven't tested it, so not my fault if it doesn't work)

```suggestion
  onLongPress = async () => {
      const { onLongPress, event, eventProperties } = this.props;
      if (!onLongPress) return;
      if (event) {
        track(event, eventProperties);
      }
      
      const res = onLongPress();
      this.setState({ pending: true });
      await res;
      this.setState({ pending: false });
  };
```

src/screens/Settings/Accounts/index.tsx
```diff
@@ -75,6 +78,15 @@ export default function AccountsSettings({ navigation }: { navigation: any }) {
             onPress={() => navigation.navigate(ScreenName.CryptoAssetsSettings)}
           />
         )}
+        {hiddenNftCollections.length > 0 && (
```
juan-cortes, line 81:
```suggestion
        {hiddenNftCollections.length && (
```

juan-cortes (approved):
Not sure about my suggestion, but otherwise looks ok to me.

PR #2453, opened 2022-04-26
LIVE-2087 - Fix - Set default timeframe from day to month

nparigi-ledger:
Set default timeframe from day to month

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-2087

Store


thomasrogerlux (approved):
Deep tech

PR #2457, opened 2022-04-28
LIVE-2083 LIVE-2084 - Reenable navigation to Lido platform app on iOS

OlivierFreyssinet:
On the Portfolio (in the Carousel) and on the account screen for Ethereum accounts, we have CTAs that lead to the Lido discover apps.

The logic for these CTAs was using deeplinking (opening this URI: `ledgerlive://discover/lido`) and the deeplinks handling and redirection logic for discover apps is in part handled by the `PlatformCatalog` screen: react-navigation navigates to the `PlatformCatalog` screen with the correct params, then that screen has a useEffect to check the validity of the params (it checks that the app manifest exists and is not filtered) and then navigate to the `PlatformApp` screen.

The issue with that is that on iOS we have removed the `PlatformCatalog` completely from the navigation logic so this deeplinking logic cannot work. That is why these CTAs have been disabled on iOS.

The simple solution then to reenable these is to implement a direct navigation instead of a deeplink logic.

#### Wallet > "Recommended" section > Lido card

> https://user-images.githubusercontent.com/91890529/165715963-7d046a67-c447-4721-9744-1a7149ed8e04.mp4

#### Ethereum account page > "Stake" CTA

> https://user-images.githubusercontent.com/91890529/165715977-dd179e78-c6a3-43ff-99dc-5b248aaecffe.mp4

### Type

Reenabling existing feature

### Context

[LIVE-2083]
[LIVE-2084]

- Wallet > "Recommended" section > Lido card
- Ethereum account page > "Stake" CTA

[LIVE-2083]: https://ledgerhq.atlassian.net/browse/LIVE-2083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIVE-2084]: https://ledgerhq.atlassian.net/browse/LIVE-2084

### DIFF
--- a/src/components/Carousel/shared.tsx
+++ b/src/components/Carousel/shared.tsx
@@ -96,6 +96,7 @@ const LidoSlide: SlideProps = {
   onPress: navigate => {
     navigate(ScreenName.PlatformApp, {
       platform: "lido",
+      name: "Lido",
     });
   },
   name: "Lido",
--- a/src/components/Carousel/shared.tsx
+++ b/src/components/Carousel/shared.tsx
@@ -1,9 +1,9 @@
 import React from "react";
-import { Platform } from "react-native";
 import { urls } from "../../config/urls";
-import Slide from "./Slide";
+import Slide, { SlideProps } from "./Slide";
 import Illustration from "../../images/illustration/Illustration";
 import { width } from "../../helpers/normalizeSize";
+import { ScreenName } from "../../const";
 
 const illustrations = {
   dark: {
@@ -27,7 +27,7 @@ const illustrations = {
 };
 
 /*
-const AcademySlide = {
+const AcademySlide: SlideProps = {
   url: urls.banners.ledgerAcademy,
   name: "takeTour",
   title: "carousel.banners.tour.title",
@@ -49,7 +49,7 @@ const AcademySlide = {
 };
 */
 
-const BuySlide = {
+const BuySlide: SlideProps = {
   url: "ledgerlive://buy",
   name: "buyCrypto",
   title: "carousel.banners.buyCrypto.title",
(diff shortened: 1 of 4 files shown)


src/components/Carousel/shared.tsx
```diff
@@ -91,8 +91,14 @@ const SwapSlide = {
   },
 };
 
-const LidoSlide = {
+const LidoSlide: SlideProps = {
   url: "ledgerlive://discover/lido",
```
ofreyssinet-ledger, line 95: we keep this as it's used for analytics

src/screens/Platform/App.js
```diff
@@ -1,17 +1,21 @@
...
 import WebPlatformPlayer from "../../components/WebPlatformPlayer";
 
 const PlatformApp = ({ route }: StackScreenProps) => {
   const { dark } = useTheme();
   const { platform: appId, ...params } = route.params;
+  const { setParams } = useNavigation();
   const { manifests } = usePlatformApp();
   const manifest = manifests.get(appId);
+  useEffect(() => {
+    manifest.name && setParams({ name: manifest.name });
```
ofreyssinet-ledger, line 17: When navigating to this screen, we don't necessarily have the manifest.name as we just need the `id` and not the manifest itself, so here we ensure that the title (`name`) displayed at the top of the screen is correct.

thomasrogerlux (approved):

ychen-ledger (approved):
Validated with: 
- LLM 3.0.8
- iPhone 11 / iOS 14.7
- ETH account-stake button/Lido Recommended card

PR #2463, opened 2022-04-29
Sentry integration – iteration 2

gre:
- [x] exclude more errors (all errors related to bad network, bad device conditions,.. are filtered out – so we can focus on actual LL errors)
- [ ] Fix iOS sourcemap


gre:
I'm sending the exclusion part in https://github.com/LedgerHQ/ledger-live-mobile/pull/2500 because the sourcemap isn't fixed yet :'(

PR #2467, opened 2022-04-29
LIVE-2028 - Fix - Use hide empty sub account setting on accounts list

nparigi-ledger:
Use hide empty sub account setting on accounts list

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-2028

Accounts

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -1877,7 +1877,7 @@
         "bullet2": "Ledger Live version, OS region, language and region"
       },
       "hideEmptyTokenAccounts": "Hide empty token accounts",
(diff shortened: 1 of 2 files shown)


ofreyssinet-ledger:
@nparigi-ledger I see 2 things on the ticket that don't seem to be reflected in the code changes:
- this "hiding" of accounts should be determined by some user settings (and maybe it should be hidden by default, idk if that has to be changed)
- there is a mention of some copy to update

OlivierFreyssinet (approved):
LGTM, tested and approved 👍

PR #2472, opened 2022-05-02
LIVE-2214 - Disable uppercase when displaying an address in DeviceAction

ofreyssinet-ledger:
In buy flow (an possibly in other flows), the user has to check that the address displayed on device matches the one displayed in-app. The issue is that in the app it's written with uppercase.

With this fix we remove the "uppercase" text transformation that was applied in the component displaying the address, so the address will be displayed with its original casing.

### Type

Bug fix

### Context

[LIVE-2214]

[LIVE-2214]: https://ledgerhq.atlassian.net/browse/LIVE-2214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/DeviceAction/rendering.tsx
+++ b/src/components/DeviceAction/rendering.tsx
@@ -67,9 +67,19 @@ const TitleContainer = styled(Flex).attrs({
   py: 8,
 })``;
 
-const TitleText = ({ children }: { children: React.ReactNode }) => (
+const TitleText = ({
+  children,
+  disableUppercase,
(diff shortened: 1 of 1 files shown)


thomasrogerlux (approved):

ggilchrist-ledger:
tested on Android (samsung) and everything looks fine 👍

PR #2477, opened 2022-05-03
Feature/rating

cgrellard-ledger:
Jira : [LIVE-1857]
Figma : https://www.figma.com/file/CTMrn7rPGH08SPhB7PSiot/LLM-%2F-Rating-app?node-id=2870%3A28734

Need the live-common package to be updated

[LIVE-1860]: https://ledgerhq.atlassian.net/browse/LIVE-1860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://user-images.githubusercontent.com/17146928/166492347-d76b9b2e-e1fb-4c90-9005-ff45a2b51e1c.mp4

https://user-images.githubusercontent.com/17146928/166492364-fa4a3858-675e-4c6e-943d-564833d19b33.mp4

[LIVE-1857]: https://ledgerhq.atlassian.net/browse/LIVE-1857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ofreyssinet-ledger:
is it possible to do a rebase on `develop` and skip the "merge develop into this branch" commit so we have a clean diff & history ?

src/index.js
```diff
@@ -193,6 +196,7 @@ function App({ importDataString }: AppProps) {
 
       <AnalyticsConsole />
       <ThemeDebug />
+      {ratings?.enabled ? <RatingsModal /> : null}
```
thomasrogerlux, line 199: This works but you can also use `<FeatureToggle />` for simple logic like that

src/screens/Settings/index.tsx
```diff
@@ -1,24 +1,33 @@
...
 type Props = {
   navigation: any;
 };
 
 export default function Settings({ navigation }: Props) {
+  const dispatch = useDispatch();
   const { t } = useTranslation();
   const accounts = useSelector(accountsSelector);
+  const [, setRatingsModalOpen] = useRatings();
+  const ratings = useFeature("learn"); // TODO : replace learn with ratings
```
thomasrogerlux: Todo police

thomasrogerlux (changes requested):

src/screens/Settings/index.tsx
thomasrogerlux: Ah you fixed it while I reviewed

thomasrogerlux (approved):

cgrellard-ledger:
> is it possible to do a rebase on `develop` and skip the "merge develop into this branch" commit so we have a clean diff & history ?

On it

ofreyssinet-ledger:
would be cool for a feature like that to write a short description of how it works & what it achieves

src/screens/RatingsModal/Enjoy.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const ratings = useFeature("ratings");
+
+  const dispatch = useDispatch();
+  const ratingsDataOfUser = useSelector(ratingsDataOfUserSelector);
+  const ratingsHappyMoment = useSelector(ratingsHappyMomentSelector);
+  const goToStore = useCallback(() => {
+    track("RedirectedToStore", { source: ratingsHappyMoment.route_name });
+    Linking.openURL(
+      Platform.OS === "ios" ? urls.applestoreRate : urls.playstore,
+    );
```
ofreyssinet-ledger, line 38: It would be cool to directly leverage the platform's native APIs to do that in-app (not critical, maybe we can improve on this later)
https://github.com/oblador/react-native-store-review

src/reducers/ratings.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+/* eslint import/no-cycle: 0 */
+import { handleActions } from "redux-actions";
+import type { State } from ".";
+
+export type RatingsState = {
```
ofreyssinet-ledger, line 6: It would be good to add a short description above each property of the state.
e.g: 
```
{
  /** here explain what this property is for */
  currentRoute?: string;
}
```

src/reducers/ratings.js
```diff
@@ -0,0 +1,61 @@
...
+  isRatingsModalOpen: boolean,
+  currentRoute?: string,
+  happyMoment?: any,
+  dataOfUser?: any,
```
ofreyssinet-ledger, line 10: is it possible to have better typing here and on the property above?

src/logic/ratings.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+export async function setRatingsDataOfUserInStorage(ratingsDataOfUser) {
+  await AsyncStorage.setItem(
+    ratingsDataOfUserAsyncStorageKey,
+    JSON.stringify(ratingsDataOfUser),
+  );
+}
+
+const useRatings = () => {
+  const ratings = useFeature("ratings");
```
ofreyssinet-ledger, line 54: maybe rename this to `ratingsFeature` or `ratingsConfig` ?
Also it would be great to have a typing more specific than `any` as this object seems to contain a variety of pretty different stuff

src/screens/RatingsModal/Enjoy.tsx
cgrellard-ledger: We thought about that and did some discovery, this would bring some issues and the native flow didn't met our expectations, we'll be using it later when the overall quality of the app is better

src/logic/ratings.tsx
```diff
@@ -0,0 +1,204 @@
...
+} from "../reducers/ratings";
+import {
+  setRatingsModalOpen,
+  setRatingsCurrentRoute,
+  setRatingsHappyMoment,
+} from "../actions/ratings";
+import { languageSelector } from "../reducers/settings";
+import { track } from "../analytics";
+
+const ratingsDataOfUserAsyncStorageKey = "ratingsDataOfUser";
```
ofreyssinet-ledger, line 23: you could use the settings for simple storage things like that instead of reimplementing a storage logic (I could be wrong as i'm still reading the code and maybe I don't have all the context necessary)

cgrellard-ledger: I made a separate logic as the ratings are not strictly related to the settings and it is a very distinct and independent feature. I'm open to discuss this topic more in depth to make the best decision

src/screens/RatingsModal/Init.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const dispatch = useDispatch();
+  const ratingsDataOfUser = useSelector(ratingsDataOfUserSelector);
+  const ratingsHappyMoment = useSelector(ratingsHappyMomentSelector);
+  const goToEnjoy = useCallback(() => {
+    setStep("enjoy");
+    track("Satisfied", { source: ratingsHappyMoment.route_name });
+  }, [ratingsHappyMoment.route_name, setStep]);
+  const goToDisappointed = useCallback(() => {
+    setStep("disappointed");
+    track("Disappointed", { source: ratingsHappyMoment.route_name });
```
ofreyssinet-ledger, line 39: I feel like this **logic** should be centralized in `useRatings()`log), so rather than having `useRatings()` returning `[isRatingsModalOpen, setRatingsModalOpenCallback]` it would return an object and a bunch of functions that are supposed to handle the logic (for instance here it would be something like `handleDisappointed`). That way the entire logic of the ratings is in one place rather than being widespread in components which in my opinion should only have visibility on the UI state and pseudo navigation (setStep).

That will also allow you to avoid duplicating code and have everything nicely factored (for instance the repetitive calls to
```
dispatch(setRatingsDataOfUser(ratingsDataOfUserUpdated));
setRatingsDataOfUserInStorage(ratingsDataOfUserUpdated);
```

ofreyssinet-ledger: well actually then you would have to remove the `useEffect()` (that handles navigation changes) from `useRatings()` (and remove all other side effects from it)

cgrellard-ledger: Totally agree. Thanks a lot for the feedback !

src/screens/RatingsModal/DisappointedForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useSelector } from "react-redux";
+import useFeature from "@ledgerhq/live-common/lib/featureFlags/useFeature";
+import { Flex } from "@ledgerhq/native-ui";
+import { WebView } from "react-native-webview";
+import styled from "styled-components/native";
+import { track } from "../../analytics";
+import { ratingsHappyMomentSelector } from "../../reducers/ratings";
+
+const injectedJavascript = `
+setTimeout(function() {
```
ofreyssinet-ledger, line 9: is it possible to implement a "wait for item to be there / retry logic" rather than a one shot strategy like this timeout of an arbitrary duration ? with a `setInterval` it should be pretty straightforward

src/logic/ratings.tsx
```diff
@@ -0,0 +1,316 @@
...
+  const isHappyMomentTriggered = useCallback(
+    (happyMoment: RatingsHappyMoment, ratingsNewRoute?: string) =>
+      (happyMoment.type === "on_enter" &&
+        happyMoment.route_name === ratingsNewRoute) ||
+      (happyMoment.type === "on_leave" &&
+        happyMoment.route_name === ratingsOldRoute),
+    [ratingsOldRoute],
+  );
+
+  const triggerRouteChange = useCallback(
```
ofreyssinet-ledger, line 179: maybe here rename to `handleRouteChanged` or `onRouteChanged` no?

src/screens/RatingsModal/Enjoy.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const Enjoy = ({ closeModal }: Props) => {
+  const {
+    ratingsHappyMoment,
+    handleEnjoyNotNow,
+    handleGoToStore,
+  } = useRatings();
+  const goToStore = useCallback(() => {
+    track("RedirectedToStore", { source: ratingsHappyMoment.route_name });
+    Linking.openURL(
```
ofreyssinet-ledger, line 28: You could also put this logic in `handleGoToStore`

OlivierFreyssinet (changes requested):
Great work! 👍 
Just a small change needed in the JS injected in the WebView to make sure that we catch the submit button press event, apart from that no apparent problem.

src/screens/RatingsModal/DisappointedForm.tsx
```diff
@@ -2,16 +2,20 @@ import React, { useCallback } from "react";
...
 import { WebView } from "react-native-webview";
 import styled from "styled-components/native";
-import { track } from "../../analytics";
+import { track, TrackScreen } from "../../analytics";
 import useRatings from "../../logic/ratings";
 
 const injectedJavascript = `
-setTimeout(function() {
+const interval = setInterval(addListenerOnFormSubmitButton, 100);
+function addListenerOnFormSubmitButton() {
```
ofreyssinet-ledger, line 10: noice 👌

OlivierFreyssinet (approved):
noice

PR #2481, opened 2022-05-04
chore: add react-navigation flipper devtools

Justkant:
Adds developer tools to make debugging easier when using React Navigation.
https://reactnavigation.org/docs/devtools/

### Type

Devtools

### Context

While working on #2462 I needed to debug the navigation and this helped me a lot

- none (it's only available while developing the app and using flipper)


OlivierFreyssinet (approved):
great! 👍 
does it work on android too?

Justkant:
> great! 👍 does it work on android too?

I only tested on Android 😅

PR #2211, opened 2022-02-16
Poc detox

ggilchrist-ledger:
PR to add initial tests for LLM, add iOS tests into the CI and do a bit of refactoring from @JunichiSugiura 's initial work on Detox. The two tests are for a basic Nano X Onboarding flow and a basic Add Password flow. The CI will then be able to run these tests for iOS as part of the `Detox E2E CI` workflow. 

Not included in PR:
- Android CI tests. Still some issues to iron out on this
- Fast CI tests. They take about 1 hour to run (mostly due to the app building).

https://user-images.githubusercontent.com/87659842/166466076-0ab06796-9264-4ca5-ac0d-01f458c640e3.mov

https://user-images.githubusercontent.com/87659842/166466127-3e76c542-eba0-43e1-b420-02eeadad55e6.mov

### Type

- Development of Mobile E2E Test Framework 
- Addition of iOS CI test workflow for nightly testing

### Context

[LLM automation project](https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/3529048065/PROJECT+Mobile+Test+Framework)

### Parts of the app affected / Test plan

To test:
- Have your local environment setup for Android/iOS React Native development
- It's a good idea to clean and `yarn` the repo first
- Build the app. For example: `yarn detox build --configuration android.debug` or `yarn detox build --configuration ios.staging`
- Run the tests (if you are running a debug build, you will need to have the metro bundler running in another terminal in the directory with `yarn start`): `yarn detox test -c android.debug --loglevel verbose` or `yarn detox test -c ios.staging --loglevel verbose`

PR #2484, opened 2022-05-04
Poc detox

valpinkman:
copy of https://github.com/LedgerHQ/ledger-live-mobile/pull/2211


src/screens/Onboarding/steps/setupDevice/scenes/PairNew.tsx
```diff
@@ -30,7 +30,12 @@ const Next = ({
   const { t } = useTranslation();
 
   return (
-    <Button type="main" size="large" onPress={onNext}>
+    <Button
+      type="main"
+      size="large"
+      onPress={onNext}
+      testID="Onboarding - Pair New Nano"
```
juan-cortes, line 37: Won't this give you troubles? Ids shouldn't have spaces

e2e/models/onboarding/onboardingSteps.js
```diff
@@ -0,0 +1,62 @@
+import * as bridge from "../../bridge/server";
+import * as testHelpers from "../../helpers";
+
+export default class OnboardingSteps {
+  static async waitForPageToBeVisible() {
+    await testHelpers.waitForElementByText("Get started");
+  }
+
+  static async startOnboarding() {
+    await testHelpers.tapByText("Get started");
```
juan-cortes, line 10: Won't all these break if we change any of the wordings? Why not rely on ids like on other pages? 🤔

e2e/specs/password.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    await loadConfig("1AccountBTC1AccountETH", true);
+
+    await PortfolioPage.waitForPageToBeVisible();
+    await PortfolioPage.navigateToSettings();
+    await SettingsPage.navigateToGeneralSettings();
+    await GeneralSettingsPage.togglePassword();
+    await GeneralSettingsPage.enterNewPassword(CORRECT_PASSWORD);
+    await GeneralSettingsPage.enterNewPassword(CORRECT_PASSWORD); // confirm password step
+    await device.sendToHome(); // leave LLM app and go to phone's home screen
+    await delay(60001); // password takes 60 seconds of app inactivity to activate
```
juan-cortes, line 25: You could override https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/constants.js#L21 if in a test run to not have to wait I guess.

juan-cortes (approved):
Few remarks, nothing blocking a merge for me

src/screens/Onboarding/steps/setupDevice/scenes/PairNew.tsx
valpinkman: Fixed

e2e/models/onboarding/onboardingSteps.js
Justkant: I would say that test-ids are useful but don't test in the way the user would, some more infos: https://testing-library.com/docs/queries/about/#priority

PR #2496, opened 2022-05-06
LIVE-2033 - Bugfix: Nft fix titles background

nparigi-ledger:
Fix multiple nft screens background that were using old theme

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LIVE-2033

Account/NFT


package.json
```diff
@@ -3,7 +3,7 @@
     "node": ">=14"
   },
   "name": "ledgerlivemobile",
-  "version": "3.1.0",
+  "version": "3.1.0-rc1",
```
LFBarreto, line 6: could you revert just this :pray:

LFBarreto (changes requested):

PR #2503, opened 2022-05-11
[LIVE-2320] - Bugfix: Fix missing import for Clipboard

lambertkevin:
[LIVE-2320](https://ledgerhq.atlassian.net/browse/LIVE-2320)

⚠️ FYI: When tested in the iOS simulator, it's still crashing the app... I had to use a real device to actually test it without crash.

### Type

Bugfix

### Context

https://user-images.githubusercontent.com/44363395/167811468-4b718991-fb96-4fea-8798-ad6955796acd.mov

https://user-images.githubusercontent.com/44363395/167815191-9446792e-819f-4725-977e-58ba274ecf13.mp4

Go to NFT Viewer & click on any "copy" button

### DIFF
--- a/src/components/Nft/NftViewer.tsx
+++ b/src/components/Nft/NftViewer.tsx
@@ -17,6 +17,7 @@ import { BigNumber } from "bignumber.js";
 import { useSelector } from "react-redux";
 import { Button, Icons } from "@ledgerhq/native-ui";
 import { useTranslation, Trans } from "react-i18next";
+import Clipboard from "@react-native-community/clipboard";
 import { ProtoNFT } from "@ledgerhq/live-common/lib/types";
 import { useNavigation, useTheme } from "@react-navigation/native";
 import { getAccountBridge } from "@ledgerhq/live-common/lib/bridge";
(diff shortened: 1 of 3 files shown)


src/components/Nft/NftViewer.tsx
```diff
@@ -17,6 +17,7 @@ import { BigNumber } from "bignumber.js";
 import { useSelector } from "react-redux";
 import { Button, Icons } from "@ledgerhq/native-ui";
 import { useTranslation, Trans } from "react-i18next";
+import Clipboard from "@react-native-community/clipboard";
```
haammar-ledger, line 20: I'm wondering how it could build before without that import :thinking:

ghost (approved):
Code looks OK, pending QA approval.

src/components/Nft/NftViewer.tsx
lambertkevin: We don't have any CI anymore on TS files... So that + everything is red in VSCode cause we don't have either the proper configuration, or we didn't take the time to actually fix all of our errors, it makes it almost impossible to catch this kind of things...

ychen-ledger (approved):
Validated with: 
- LLM Fix build: 
  https://github.com/LedgerHQ/ledger-live-build/actions/runs/2307098104
   https://github.com/LedgerHQ/ledger-live-build/actions/runs/2306086788
- Android 12/ iOS 14.7

PR #2515, opened 2022-05-17
Set AdServices as optional to not crash on some phones

gre:
see https://developer.apple.com/forums/thread/673708
there is an history of this framework to causes crash at boot
I wasn't able to reproduce this myself but this PR makes it optional as the thread is suggesting to

### Type

bugfix

### Context

possible incident on users end where **some** phones crash at boot. Something we don't see yet in Apple Crashes that would be good to confirm this is the actual fix 🤞


LFBarreto (approved):

LFBarreto:
ported to https://github.com/LedgerHQ/ledger-live/pull/109

